PR #4, opened 2026-02-11
fix: return screenshot as image content instead of parsing PNG as JSON

liveinknewgithub:
## Problem

The `camofox_screenshot` tool is completely broken — every call fails.

### Bug 1: Parsing PNG binary as JSON

In `plugin.ts`, the `camofox_screenshot` tool calls:

```ts
const result = await fetchApi(baseUrl, `/tabs/${tabId}/screenshot?userId=${userId}`);
```

`fetchApi()` always calls `res.json()` to parse the response (line 181). But the server's `GET /tabs/:tabId/screenshot` endpoint returns **raw PNG binary** (`Content-Type: image/png`), not JSON. This causes a JSON parse error on every invocation.

### Bug 2: Wrong content type in tool result

Even if parsing succeeded, `toToolResult()` wraps the result as:

```ts
{ content: [{ type: "text", text: JSON.stringify(data, null, 2) }] }
```

Screenshots need to be returned as an **image content block** so the LLM can actually see the image.

## Fix

Bypass `fetchApi()` for the screenshot endpoint:

1. Use `fetch()` directly to avoid the forced `.json()` parsing
2. Read the response as `arrayBuffer()` and base64-encode the PNG
3. Return as `{ type: "image", data: base64, mimeType: "image/png" }`

## Tests

**15 new tests** (all passing):

### Unit tests (`tests/unit/screenshotToolResult.test.js`)
Uses a mock HTTP server returning real PNG binary to test both old and new code paths:
- ✅ Proves `fetchApi()` throws when parsing PNG as JSON (the bug)
- ✅ Proves `toToolResult()` wraps as `{ type: "text" }` not image
- ✅ Fixed code returns `{ type: "image", mimeType: "image/png" }` content block
- ✅ Base64 data decodes back to valid PNG (magic bytes check)
- ✅ Base64 round-trips to exact original bytes
- ✅ Result contains no text content blocks
- ✅ 404 errors propagate correctly

### E2E tests (`tests/e2e/screenshot.test.js`)
Full integration against the real Camoufox server:
- ✅ Screenshot returns valid PNG binary with correct magic bytes
- ✅ Response `Content-Type` is `image/png`
- ✅ PNG binary response is NOT valid JSON
- ✅ Base64 encoding round-trips correctly
- ✅ Full-page screenshots are larger than viewport screenshots
- ✅ Non-existent tab returns 404 JSON error
- ✅ Screenshots work after navigation
- ✅ Screenshots work after click interactions

## Changes

- **plugin.ts**: Rewrote `camofox_screenshot` execute function to handle binary PNG response correctly
- **tests/unit/screenshotToolResult.test.js**: Unit tests proving old path broken + new path correct
- **tests/e2e/screenshot.test.js**: E2E tests for the screenshot server endpoint

### DIFF
--- a/tests/e2e/screenshot.test.js
+++ b/tests/e2e/screenshot.test.js
@@ -1,0 +1,202 @@
+const { startServer, stopServer, getServerUrl } = require('../helpers/startServer');
+const { startTestSite, stopTestSite, getTestSiteUrl } = require('../helpers/testSite');
+const { createClient } = require('../helpers/client');
+
+describe('Screenshot', () => {
+  let serverUrl;
+  let testSiteUrl;
+
+  beforeAll(async () => {
+    const port = await startServer(9379);
+    serverUrl = getServerUrl();
+
+    const testPort = await startTestSite();
+    testSiteUrl = getTestSiteUrl();
+  }, 120000);
+
+  afterAll(async () => {
+    await stopTestSite();
+    await stopServer();
+  }, 30000);
+
+  test('screenshot returns raw PNG binary with correct magic bytes', async () => {
+    const client = createClient(serverUrl);
+
+    try {
+      const { tabId } = await client.createTab(`${testSiteUrl}/pageA`);
+
+      const buffer = await client.screenshot(tabId);
+
+      expect(buffer).toBeDefined();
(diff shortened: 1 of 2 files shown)


skyfallsin:
@liveinknewgithub needs tests

PR #6, opened 2026-02-11
Add cookie injection endpoint + Netscape cookie import tool

BuffMcBigHuge:
Adds cookie injection support to camofox-browser.

What's included:
- Server: `POST /sessions/:userId/cookies` to inject Playwright cookie objects into the user's BrowserContext.
- OpenClaw tool: `camofox_import_cookies` to parse a Netscape-format cookie file (optional `domainSuffix`) and POST cookies to the server.
- README: documents the new tool + endpoint.

Tested:
- `npm ci`
- `npm test`

Use case:
Authenticate against sites like LinkedIn without an interactive login flow by importing browser-exported cookies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ If you’re using the OpenClaw plugin, you can import a Netscape-format cookie f
 - Tool: `camofox_import_cookies`
 - Server endpoint: `POST /sessions/:userId/cookies`
 
+**Security:** this endpoint is disabled unless `CAMOFOX_API_KEY` is set on the server. When enabled, callers must include `Authorization: Bearer <CAMOFOX_API_KEY>`.
+
 ```bash
 # OpenClaw tool usage (conceptual)
 # camofox_import_cookies({ cookiesPath: "/path/to/cookies.txt", domainSuffix: "linkedin.com" })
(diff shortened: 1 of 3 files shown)


BuffMcBigHuge:
Thanks for the feedback @skyfallsin .

• Secured the cookie endpoint: POST /sessions/:userId/cookies is now disabled unless CAMOFOX_API_KEY is set, and when enabled it requires Authorization: Bearer <key>.
• Validated cookie objects before passing to Playwright: each cookie must have name, value, and domain (otherwise returns a clear 400 with details instead of a cryptic 500).
• Plugin updated to read process.env.CAMOFOX_API_KEY and send the Authorization header; if missing, it throws a clear error.
• README cleanup: removed the duplicate ## Usage heading and documented the auth requirement for cookie injection.
• Netscape parser tiny fix: changed == → === for the secure field check.
All tests passed locally (npm test).

skyfallsin:
Landed on main via merge commit 1826ffaa723e434a5c17095d495b2939e5a2ace3. Thanks @BuffMcBigHuge for the cookie import feature! Added security hardening (path traversal prevention, field sanitization, timing-safe auth, body limits) and test coverage (30 tests) on top.

PR #51, opened 2026-04-07
fix: bind to 0.0.0.0 and fix multi-arch Docker builds for Apple Silicon

jecruz:
## Summary

- **Bind server to `0.0.0.0`** — without an explicit host, Node.js may bind to IPv6 loopback only, silently dropping connections forwarded from the Docker bridge network
- **Fix multi-arch `TARGETARCH` mapping** — Docker buildx injects `TARGETARCH` as `amd64`/`arm64`, but Camoufox release filenames use `x86_64`/`aarch64`. The previous `ARG` interpolation produced invalid URLs (`lin.arm64.zip`, `lin.amd64.zip`) on both `linux/arm64` and `linux/amd64` when using `--platform`. A `case` statement now maps arch names correctly at build time.

## Changes

- `server.js`: `app.listen(PORT, '0.0.0.0', ...)` — explicit host binding for Docker
- `Dockerfile`: Replace `ARG CAMOUFOX_URL` interpolation with a `RUN case` statement that maps `amd64→x86_64` and `arm64→aarch64`

## Test Plan

- [ ] `docker build .` (no `--platform`) pulls `lin.x86_64.zip` — defaults correctly
- [ ] `docker buildx build --platform linux/arm64 .` pulls `lin.aarch64.zip`
- [ ] `docker buildx build --platform linux/amd64 .` pulls `lin.x86_64.zip`
- [ ] `docker run -p 3000:3000` — server reachable from host

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,45 @@
+# camofox-browser
+
+## Tech Stack
+
+- Runtime: Node.js 20
+- Framework: Express (REST API)
+- Browser Engine: Camoufox (Firefox fork with C++-level fingerprint spoofing)
+- Container: Docker (multi-arch: linux/amd64, linux/arm64)
+
+## Structure
+
+```
+server.js        # Express entry point
+lib/             # Core modules (browser, routes, utils)
+tests/           # Jest test suite
+scripts/         # Build/utility scripts
+Dockerfile       # Multi-arch image build
+docker-compose.yml
+```
+
+## Conventions
+
+- Commits: conventional commits (`fix(scope): desc`)
+- Branch: `fix/*`, `feature/*` from `main`
+- Tests: Jest (`npm test`)
+
+## Scripts
+
+- `npm start` — start server
+- `npm test` — run tests
+- `docker build` — build image
+
+## Docker Multi-Arch Notes
+
+- `TARGETARCH` is injected by Docker buildx as `amd64` or `arm64`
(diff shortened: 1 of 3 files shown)


skyfallsin:
Hey @jecruz — thanks for this. Two things we're taking from it (and you'll be credited):

- **`0.0.0.0` bind** — good catch, Node defaulting to IPv6 loopback silently breaks Docker bridge networking. We've picked this up on our `next` branch.
- **`TARGETARCH` mapping** — the `amd64→x86_64` / `arm64→aarch64` arch translation is correct and we've addressed multi-arch builds on `next` (via #51's partial merge + our own Makefile-based approach).

The Dockerfile takes a different approach from ours — we pre-download binaries with `make fetch` and bind-mount `dist/` rather than curling at build time, so we won't take those changes directly. And the server.js diff is mostly formatting (single→double quotes, trailing commas) — the substantive change is just the bind address which we've got.

Heads up — the PR includes a `CLAUDE.md` that looks like it's from your local setup. Might want to add that to `.gitignore` so it doesn't sneak into future PRs.

We won't merge this directly since the relevant pieces already landed on `next`, but appreciate the contribution — especially the arch mapping fix.

skyfallsin:
Integrated into v1.6.0 — multi-arch Dockerfile fix included. Thank you! 🎉

https://github.com/jo-inc/camofox-browser/releases/tag/v1.6.0

PR #62, opened 2026-04-15
feat: Add optional persistence

company8:
## Summary

Adds opt-in per-user browser storage state persistence (cookies + localStorage) so agent sessions survive container restarts, deploys, and idle timeouts, without re-importing cookies every time.

**Disabled by default.** Existing deployments are unaffected unless both env vars are set.

## How it works

- On session create: if a persisted `storageState` exists for the `userId`, Playwright restores it.
- On cookie import and session teardown (and SIGTERM shutdown): checkpoint cookies/localStorage to disk via atomic tmp-write + rename.
- Each `userId` maps to a deterministic SHA256-hashed subdirectory under `CAMOFOX_PROFILE_DIR`, so arbitrary userIds are path-safe.

## Configuration

| Env var | Purpose |
|---|---|
| `CAMOFOX_PERSISTENCE` | `1` / `true` to opt in |
| `CAMOFOX_PROFILE_DIR` | Root dir for persisted state (required when flag is on) |

Both must be set, either alone is a no-op. The Makefile gains a `PERSISTENCE=1` shortcut for Docker users that wires up the env vars and a volume mount.

## What's in the PR

- `lib/persistence.js` — `getUserPersistencePaths`, `loadPersistedStorageState`, `persistStorageState` (atomic writes, cleanup on failure)
- `lib/inflight.js` — `coalesceInflight` primitive so concurrent `getSession` calls for the same userId share one creation promise (prevents races that could clobber persisted state)
- `lib/config.js` — `persistenceEnabled` + `profileDir` surfaced on config; env vars added to the whitelist
- `lib/cookies.js` — `importBootstrapCookies` helper extracted for testability
- `server.js` — restore on create, persist on import/destroy, persist-on-shutdown via `closeAllSessions({ persist: true })`, `getSession` wrapped in `coalesceInflight`
- Tests: 12 new unit tests across `persistence`, `inflight`, `autoCookieImport`
- README: new env vars documented, short persistence section
- Makefile: optional `PERSISTENCE=1` target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,17 @@ endif
 IMAGE        := camofox-browser:$(VERSION)-$(ARCH)
 CAMOUFOX_ZIP := dist/camoufox-$(ARCH).zip
 YTDLP_BIN    := dist/yt-dlp-$(ARCH)
+COOKIES_DIR  ?= $(HOME)/.camofox/cookies
+PROFILE_DIR  ?= $(HOME)/.camofox/profiles
+
+PERSISTENCE  ?=
+ifeq ($(filter 1 true yes on,$(PERSISTENCE)),)
+  PERSISTENCE_ENV   :=
+  PERSISTENCE_MOUNT :=
+else
+  PERSISTENCE_ENV   := -e CAMOFOX_PERSISTENCE=1 -e CAMOFOX_PROFILE_DIR="$(PROFILE_DIR)"
+  PERSISTENCE_MOUNT := -v "$(PROFILE_DIR):$(PROFILE_DIR)"
+endif
 
 CAMOUFOX_URL := https://github.com/daijro/camoufox/releases/download/v$(VERSION)-$(RELEASE)/camoufox-$(VERSION)-$(RELEASE)-lin.$(CAMOUFOX_ARCH).zip
 YTDLP_URL    := https://github.com/yt-dlp/yt-dlp/releases/latest/download/yt-dlp_linux$(YTDLP_ARCH)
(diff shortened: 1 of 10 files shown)


skyfallsin:
Hey @company8 — really nice work here. The atomic writes, the inflight coalescing fix, the closeSession consolidation — all solid.

We've been building out a plugin system on our `next` branch and after reviewing this we decided persistence fits better as a plugin than a core addition. Camofox is stateless by default and we want to keep it that way, with persistence as an opt-in layer.

Here's what we did:

- **Took your lib modules as-is** — `persistence.js`, `inflight.js`, the `importBootstrapCookies` extraction, and all the tests. Your commit, your authorship.
- **Took the `closeSession`/`closeAllSessions` refactor** — this was a big cleanup, 7 copy-pasted teardown blocks consolidated into one path. We adapted it onto our `next` branch which has some extra plugin event emissions.
- **Wrapped the actual persistence behavior as `plugins/persistence/`** — it hooks `session:creating`, `session:created`, `session:cookies:import`, `session:destroyed`, and `server:shutdown` on the event bus. No core server.js changes needed beyond the closeSession refactor.
- **Moved config to per-plugin config** in `camofox.config.json` instead of env vars in core config:
  ```json
  { "plugins": { "persistence": { "enabled": true, "profileDir": "/data/profiles" } } }
  ```
  (`CAMOFOX_PROFILE_DIR` env var still works as an override.)

The `coalesceInflight` fix for concurrent `getSession()` races is genuinely useful independent of persistence — good catch.

We won't merge this PR directly since the code landed via cherry-pick + rework on `next`, but you'll show up in the contributor graph (author on the libs commit, co-authored-by on the others). Thanks for the contribution — this was one of the more thoughtful PRs we've gotten.

skyfallsin:
Integrated into v1.6.0 — persistence ships as a built-in plugin, enabled by default. Thank you! 🎉

https://github.com/jo-inc/camofox-browser/releases/tag/v1.6.0

PR #63, opened 2026-04-15
fix: clean up orphaned camoufox temp files on startup

### LINKED ISSUE #58: Browser crashes leak .so temp files in /tmp (115GB accumulated)
## Bug

When browser instances crash or are force-killed, `.fea5*.so` temp files (~4.3MB each) are left behind in `/tmp`. These are hidden files (dot-prefixed) and accumulate silently.

In our production environment, **27,852 leaked files totaled 115.7 GB**, filling the disk and crashing the host agent.

## Environment
- **Package:** `@askjo/camoufox-browser@1.0.12`
- **OS:** Linux (Ubuntu)
- **Camoufox engine:** cached in `~/.cache/camoufox/`
- **Usage:** Auto-started as an OpenClaw/Hermes plugin on port 9377

## Reproduction
1. Start camoufox-browser server
2. AI agent opens browser sessions via the tool
3. Agent crashes or is killed (SIGTERM/SIGKILL) before browser.close() runs
4. `.fea5*.so` files remain in `/tmp` — each ~4.3MB
5. Over days/weeks, disk fills completely

## Diagnosis
```bash
# Hidden temp files not visible with normal ls
ls -la /tmp/ | grep '.fea5'

# Count and size
find /tmp -name '.fea5*.so' | wc -l       # 27,852
du -sh /tmp/.fea5*.so                      # 115.7 GB
```

Also present: `.5ef7*.node` files (mostly 0 bytes, same pattern).

## Root Cause

Browser instances are spawned with temp `.so` files but `close()` is never called when the process exits abnormally. No `try/finally` blocks, signal handlers, or temp file cleanup on startup.

## Suggested Fix

1. **Add `try/finally` cleanup** around browser lifecycle so temp files are always removed:
   ```js
   try {
     browser = await camoufox.launch(options);
     // ... use browser
   } finally {
     await browser.close();
   }
   ```

2. **Add signal handlers** (`SIGTERM`, `SIGINT`) that call `browser.close()` before exiting

3. **Stale temp file cleanup on startup** — scan `/tmp` for orphaned `.fea5*.so` / `.5ef7*.node` files and remove them

4. **Use a dedicated temp directory** (e.g., `/tmp/camoufox-{pid}/`) so cleanup can be per-process

## Workaround (consumer-side)

Cron job to periodically clean up:
```bash
find /tmp -name '.fea5*.so' -mtime +1 -delete
find /tmp -name '.5ef7*.node' -mtime +1 -delete
```

mvanhorn:
Fixes #58.

When the server is killed with SIGKILL or the kernel OOMs the process, userland cleanup never runs and camoufox's internal `.fea5*.so` and `.5ef7*.node` temp files survive in `os.tmpdir()`. They accumulate. The reporter saw 27,852 files totaling 115 GB on a production host before the disk filled.

`SIGTERM` / `SIGINT` already trigger `gracefulShutdown` -> `browser.close()`, which removes the files via camoufox's own cleanup. The only gap is the non-graceful exit path, which can't be fixed in userland (SIGKILL is uncatchable). A startup sweep is the right place to reclaim the leftovers.

## Changes

- `lib/tmp-cleanup.js`. Exports `cleanupOrphanedTempFiles({ tmpDir, minAgeMs, now })`. Pure Node: `fs.readdirSync` + `fs.statSync` + `fs.unlinkSync`. No `process.env`, no `child_process`, no network. Scanner-safe per `AGENTS.md`.
- `server.js`. Calls it once from the `app.listen` callback at startup, before the browser pre-warm. Logs a summary line only when files are removed.
- `tests/unit/tmpCleanup.test.js`. 5 unit tests using `fs.mkdtempSync` so they don't touch the real `/tmp`.

The 5-minute age threshold prevents deleting files from a concurrently-running instance. Patterns are restricted to the two known orphan shapes (`.fea5[hex]+.so` and `.5ef7[hex]+.node`); non-matching entries are left alone.

## Not in scope

- No changes to `gracefulShutdown` (already correct for SIGTERM/SIGINT).
- No new env vars or config surface.
- No per-process temp directory (would require a change in `camoufox-js` upstream).

## Testing

`npm test` passes 172/172 unit tests (14 suites). The 5 new tests cover: removal of aged files, skipping fresh files, leaving non-matching names, graceful handling of a missing `tmpDir`, and deterministic age comparison via an injected `now`.

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/lib/tmp-cleanup.js
+++ b/lib/tmp-cleanup.js
@@ -1,0 +1,40 @@
+import fs from 'fs';
+import path from 'path';
+
+const ORPHAN_PATTERNS = [
+  /^\.fea5[a-f0-9]+\.so$/,
+  /^\.5ef7[a-f0-9]+\.node$/,
+];
+
+export function cleanupOrphanedTempFiles({ tmpDir, minAgeMs = 5 * 60 * 1000, now = Date.now() } = {}) {
+  const result = { scanned: 0, removed: 0, bytes: 0, skipped: 0 };
+  if (!tmpDir) return result;
+
+  let entries;
+  try {
+    entries = fs.readdirSync(tmpDir);
+  } catch {
+    return result;
+  }
+
+  for (const name of entries) {
+    if (!ORPHAN_PATTERNS.some((re) => re.test(name))) continue;
+    result.scanned++;
+    const full = path.join(tmpDir, name);
+    try {
+      const st = fs.statSync(full);
+      if (!st.isFile()) continue;
+      if (now - st.mtimeMs < minAgeMs) {
+        result.skipped++;
+        continue;
+      }
+      fs.unlinkSync(full);
+      result.removed++;
+      result.bytes += st.size;
+    } catch {
+      // file vanished, permission denied, or race with another process - skip silently
+    }
+  }
+
+  return result;
+}
--- a/server.js
+++ b/server.js
@@ -23,6 +23,7 @@ import {
   startMemoryReporter, stopMemoryReporter,
 } from './lib/metrics.js';
 import { actionFromReq, classifyError } from './lib/request-utils.js';
+import { cleanupOrphanedTempFiles } from './lib/tmp-cleanup.js';
 
 const CONFIG = loadConfig();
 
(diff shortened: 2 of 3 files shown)


skyfallsin:
Thanks @mvanhorn — merged into our `next` branch with your authorship preserved.

We took the cleanup logic as-is: scan `/tmp` for orphaned `.fea5*.so` and `.5ef7*.node` files on startup, skip anything younger than 5 minutes (concurrent-instance guard), and log what was removed. Clean and well-tested.

One small addition on our side: we extracted it into `lib/tmp-cleanup.js` as a standalone module so plugins can also trigger cleanup if needed.

mvanhorn:
Thanks @skyfallsin — glad it landed cleanly. The `lib/tmp-cleanup.js` extraction is a nice call, having plugins trigger it on demand makes sense for longer-running sessions where a proc crash could leave artifacts mid-flight.

PR #66, opened 2026-04-17
feat: unified /type endpoint with keyboard mode (Ember/contenteditable support)

leoneparise:
## Summary

Extends `POST /tabs/:tabId/type` (and the openclaw `/act` `type` kind) with a `mode` parameter so the same endpoint handles both Playwright `fill()` and char-by-char keyboard typing. Needed for Ember/React contenteditable forms that reject DOM-mutation input.

### The problem

LinkedIn's DM compose form (`div.msg-form__contenteditable`) is Ember-managed and tracks message text via `beforeinput`/`InputEvent` provenance. Two failure modes when automating with the current `/type`:

1. `locator.fill()` sets text in the DOM but Ember does not register it as user input. No Send button appears. `keyboard.press("Enter")` is a no-op.
2. `document.execCommand("insertText", ...)` — text appears, the `is-active` class flips on, but still no Send button, and synthetic Enter events are ignored.

Only OS-level key events from char-by-char `page.keyboard.type()` satisfy Ember. This is a general pattern — same issue hits most React/Ember rich-text composers (Slack-style mention pickers, Notion-ish editors, etc.).

### API

```
POST /tabs/:tabId/type
{
  userId, ref?, selector?, text,
  mode?: "fill" | "keyboard",   // default "fill"
  delay?: number,               // ms/char in keyboard mode, default 30
  submit?: boolean,             // press Enter after typing, default false
  pressEnter?: boolean          // alias for submit
}
```

- `mode: "fill"` (default) — unchanged behavior: `locator.fill()` or `page.fill()`.
- `mode: "keyboard"` — focuses the target (if ref/selector given) then `page.keyboard.type(text, { delay })`. If neither ref nor selector is provided, types into whatever currently has focus.
- `submit`/`pressEnter` — presses Enter after typing inside the same lock. Saves a roundtrip for click-to-send flows and prevents races where the editor loses focus between calls.

Same additions applied to the openclaw `/act` `type` kind.

### Compatibility

- 100% back-compat. Callers not passing `mode` get identical behavior to before.
- Errors for bad inputs are explicit: `"mode must be 'fill' or 'keyboard'"`, `"ref or selector required for mode=fill"`, `"text is required"`.

### Tested

Smoke tests against Google (plain `<textarea>`):
- `mode:"fill"` (default) → text appears, matches.
- `mode:"keyboard"` with selector → text appears, matches.
- `mode:"keyboard"` with no selector, typing into focused element → text appended.
- `mode:"bogus"` → 400 error.
- `mode:"fill"` without ref/selector → 400 error.

End-to-end on LinkedIn DM compose:
```bash
curl -X POST http://localhost:9377/tabs/$TAB/type -d '{
  "userId":"agent1",
  "selector":"div.msg-form__contenteditable",
  "text":"Hey, sending via keyboard mode.",
  "mode":"keyboard",
  "submit":true
}'
```
Result: editor empties, message lands in thread, last `.msg-s-event-listitem` contains the text. `fill` mode on the same selector leaves the Send button absent.

### Alternatives considered

- **Separate `/keyboard_type` endpoint**: duplicates ref-resolution, auto-refresh, tab-lock, and error-handling plumbing. Rejected in favor of one endpoint with a mode flag.
- **Auto-detect contenteditable and switch modes**: brittle. Explicit `mode` keeps callers in control.

### DIFF
--- a/server.js
+++ b/server.js
@@ -2187,7 +2187,7 @@ app.post('/tabs/:tabId/type', async (req, res) => {
   const tabId = req.params.tabId;
   
   try {
-    const { userId, ref, selector, text } = req.body;
+    const { userId, ref, selector, text, mode = 'fill', delay = 30, submit = false, pressEnter = false } = req.body;
     const session = sessions.get(normalizeUserId(userId));
     const found = session && findTab(session, tabId);
     if (!found) return res.status(404).json({ error: 'Tab not found' });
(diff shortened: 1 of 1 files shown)


skyfallsin:
Thanks @leoneparise — another solid one. Merged into our `next` branch with your authorship preserved.

**Taken as-is:**
- `mode: 'keyboard'` for `/type` endpoint — `page.keyboard.type()` with configurable delay, essential for Ember/contenteditable fields where `fill()` doesn't trigger the right events
- Same keyboard mode support in `/act` type action
- Validation: mode must be `'fill'` or `'keyboard'`, fill requires ref/selector, keyboard allows typing into current focus
- `submit || pressEnter` unification
- Default 30ms delay

We added unit tests for the validation and mode behavior in `tests/unit/typeKeyboardMode.test.js`.

PR #68, opened 2026-04-19
feat: opt-in session tracing via Playwright

mvanhorn:
## Summary

Adds opt-in per-session [Playwright tracing](https://playwright.dev/docs/trace-viewer) on top of the v1.6.0 plugin system. Pass `trace: true` when opening the first tab and the server captures screenshots, DOM snapshots, and network activity into a single `.zip` you can open with `npx playwright show-trace`.

Three new endpoints, zero new runtime dependencies, zero behavior change for existing users.

## Demo

https://github.com/mvanhorn/camofox-browser/releases/download/evidence-session-tracing/evidence-video.mp4

Session tracing API flow

A sample trace.zip from the dogfood session is attached to the same bundle: [sample-trace.zip](https://github.com/mvanhorn/camofox-browser/releases/download/evidence-session-tracing/sample-trace.zip) - download it and run `npx playwright show-trace sample-trace.zip` to inspect exactly what the endpoint produces.

## Why tracing instead of video

Every major competitor in this space ships session capture of some kind - Steel Browser rebuilt its replay stack around MP4 in [Launch Week v2](https://steel.dev/blog/launch-week-v2-recap), Browserbase leads its [observability docs](https://docs.browserbase.com/features/observability) with video replay, AWS Bedrock AgentCore stores recordings in S3, and Cloudflare Browser Run advertises record/replay as a core feature. camofox-browser ships structured logs and single-shot screenshots, which isn't enough to reconstruct what the agent actually saw during a multi-step run.

Playwright's built-in `recordVideo` was the obvious first choice. It turns out `recordVideo` uses the Chrome DevTools Protocol `Browser.setScreencastOptions` method, which Firefox doesn't implement. On Camoufox it fails with `Protocol error (Browser.setScreencastOptions): ERROR: method 'Browser.setScreencastOptions' is not supported`. So the PR uses `context.tracing.start` / `tracing.stop` instead, which does work on Firefox and actually gives you more than a video: screenshots plus DOM snapshots plus network plus JS stacks plus console output, all in one zip that opens in Playwright's standard Trace Viewer.

## Changes

**New: `lib/tracing.js`** (122 lines)
- `ensureTracesDir`, `userTracesDir`, `tracePathFor`, `makeTraceFilename` - SHA-256-hashed per-user directory, timestamped filenames
- `resolveTracePath` - blocks `..`, `/`, `\`, hidden files, and any path that escapes the user's subdirectory
- `listUserTraces` - filters to `.zip`, sorts newest first
- `sweepOldTraces` - TTL + per-file size cap, runs on server startup

**Extended: `server.js`** (+97 lines)
- `POST /tabs` accepts `trace: true` in body. If the session already exists and isn't being traced, returns 409 with a clear message telling the caller to `DELETE /sessions/:userId` first.
- `getSession({ trace })` calls `context.tracing.start({ screenshots: true, snapshots: true, sources: false })` right after the context is created.
- `closeSession` calls `context.tracing.stop({ path })` before closing the context, so the zip is always flushed.
- Three new endpoints (list / stream / delete) right next to the existing `/sessions/:userId` handlers.
- The startup sweep in the existing `cleanupOrphanedTempFiles` block now also calls `sweepOldTraces`.

**Extended: `lib/config.js`** (+6 lines) - three new env vars (`CAMOFOX_TRACES_DIR`, `CAMOFOX_TRACES_MAX_BYTES`, `CAMOFOX_TRACES_TTL_HOURS`), added to the `serverEnv` whitelist per the CONTRIBUTING.md rule on explicit env var pass-through.

**README.md** - new Features bullet, new "Session Tracing" usage section with curl examples and `npx playwright show-trace` pointer, three env vars added to the reference table.

**Tests** - 9 unit tests covering directory creation, filename uniqueness, path traversal defense, listing filter + sort, TTL sweep, size-cap sweep, no-op on missing dir, and non-zip exclusion.

## Testing

```
Tests:       261 passed, 261 total
```

## Dogfooded

Ran the server locally with `CAMOFOX_TRACES_DIR=/tmp/demo/traces PORT=9378`, exercised the happy path end-to-end:

```
$ curl -X POST :9378/tabs -d '{"userId":"demo","sessionKey":"s1","url":"https://example.com","trace":true}'
  {"tabId":"d6a5d3a9-...","url":"https://example.com/"}

$ curl ":9378/tabs/$TAB/snapshot?userId=demo"
  - heading "Example Domain" [level=1]
  - link "Learn more" [e1]: https://iana.org/domains/example

$ curl -X POST :9378/tabs/$TAB/navigate -d '{"userId":"demo","url":"https://httpbin.org/html"}'
  {"ok":true,...}

$ curl -X DELETE :9378/sessions/demo       # flushes the trace
  {"ok":true}

$ curl :9378/sessions/demo/traces
  { "traces": [ { "filename": "trace-2026-04-19T04-32-46-041Z-b6af77.zip",
                  "sizeBytes": 12626, ... } ] }

$ curl -o session.zip :9378/sessions/demo/traces/trace-2026-04-19T04-32-46-041Z-b6af77.zip
$ unzip -l session.zip
    42129   trace.trace
     4108   trace.network
     7360   resources/page@...jpeg
      528   resources/....html
     3741   resources/....html
    16917   trace.stacks
```

Also verified:
- Path traversal rejected (HTTP 400) on `..%2F..%2Fetc%2Fpasswd`
- `trace: true` on an existing non-traced session returns HTTP 409 with a clear message
- Recordings show up in `~/.camofox/traces/<16-char-hash>/trace-<iso-timestamp>-<hex>.zip`
- `npx playwright show-trace` opens the zip in the standard viewer

## Notes for reviewers

Kept the surface small on purpose:
- Default off. `trace: true` is opt-in on `POST /tabs`.
- One flag per session, set at session creation. No mid-session toggling (documented, and guarded with a 409).
- Hashed user-id directory so the on-disk layout matches the persistence plugin's convention (`profiles/<hashed-userId>/`).
- TTL + size-cap sweep reuses the existing startup cleanup hook, not a new background interval.
- Uses only APIs that ship with Playwright today - no new `dependencies` entry in `package.json`.

Fixes no existing issue - this is a feature proposal based on ecosystem signal (Steel / Browserbase / AgentCore / Cloudflare / Laminar all ship session capture in one form or another).


skyfallsin:
Great work @mvanhorn — merged! We pushed a few improvements before landing:

- `@openapi` JSDoc annotations for the 3 trace endpoints + the `trace` field on `POST /tabs`
- Converted `listUserTraces` and the route handlers from sync to async fs (`fs/promises`) to avoid blocking the event loop
- Added `authMiddleware()` to all 3 trace endpoints — trace zips contain screenshots and DOM snapshots, so they need API key protection
- Fixed a TOCTOU race in the download endpoint: added a stream error handler so concurrent delete + download doesn't crash with an unhandled error
- 13 integration tests (including auth rejection and the TOCTOU race scenario)

Smart call using Playwright tracing over `recordVideo` given the Firefox constraint.

mvanhorn:
Thanks for the merge, @skyfallsin.

PR #69, opened 2026-04-19
feat: OpenAPI 3.0 spec at /openapi.json + HTML docs at /docs

mvanhorn:
## Summary

Exposes a hand-written OpenAPI 3.0 document at `GET /openapi.json` and a Redoc-rendered HTML viewer at `GET /docs`. The spec documents every core route in `server.js` (32 endpoints after this lands, up from 30 before). No new runtime dependencies, no behavior change for any existing endpoint.

## Demo

https://github.com/mvanhorn/camofox-browser/releases/download/evidence-openapi/evidence-video.mp4

OpenAPI spec at /openapi.json

## Why

The project has 31 REST endpoints on master but nothing machine-readable describes them. The `/` endpoint returns a four-field health blob. The only way to discover routes today is to read `server.js` line by line (I used `grep -nE "app\.(get|post|put|delete)\(" server.js` and got 30+ matches across ~2900 lines).

Steel Browser ships the same pattern at `/documentation` and their Python and Node SDKs are generated directly from that spec. Once this lands, anyone can produce SDKs for Python, Ruby, Go, and anything else, and tooling like Postman, MCP bridges, or OpenAI function-calling integrations can consume the spec directly.

## What's here

- `lib/openapi-spec.js` (~520 lines) - A single exported OpenAPI 3.0.3 object. Metadata, tags, security schemes, 20 reusable component schemas, 32 paths with parameters / requestBodies / responses.
- `server.js` (+25 lines) - Two route handlers right above the existing `GET /` handler. `REDOC_HTML` is a 13-line string that loads Redoc via CDN.
- `tests/unit/openapi-spec.test.js` (~90 lines) - The load-bearing piece. See "Drift detection" below.
- `README.md` (+3 lines) - One Features bullet + one Quick Start line pointing at `/docs`.

## Drift detection

The hand-written spec is only useful if it stays accurate. Two tests (`every server.js route appears in the spec` and `no stale routes in the spec that are not in server.js`) parse `server.js` with a regex for `app.<method>(<path>)` and cross-reference against `openApiSpec.paths`. They fail if:

- A new server.js route is added without documenting it
- A route is removed but still in the spec
- A path operation has no responses
- A request body `$ref` points to an undefined schema

Five other tests verify spec validity: OpenAPI 3.0.x shape, `info.version` matches `package.json`, the builder is pure, every operation has responses, every `$ref` resolves.

```
Tests:       259 passed, 259 total
```

## Notes for reviewers

- **Plugin-contributed routes are not in the spec.** The `youtube` plugin adds `/youtube/transcript`, the `vnc` plugin adds `/sessions/:userId/storage_state`. Those depend on which bundled plugins are enabled at runtime, so the core spec stays deterministic. A follow-up PR could have plugins contribute to the spec at registration time.
- **No new runtime deps.** Redoc is served from `cdn.redocly.com` in the HTML. If you'd rather self-host (which would matter for air-gapped deploys), happy to switch to inlining the ~200KB bundle or vendoring it.
- **Tags are organised for Redoc's sidebar:** System, Tabs, Interaction, Content, Sessions, Browser, Legacy. The `Legacy` bucket holds the pre-tabId `/act`, `/start`, `/stop`, `/navigate`, `/snapshot` endpoints so their backward-compat status is visible in the docs.
- **Security schemes** declare `BearerAuth` matching `CAMOFOX_API_KEY`. Currently only `/sessions/:userId/cookies` and `/stop` use it; the rest are unauthenticated, matching current server behavior.

Dogfooded locally: ran `PORT=9378 node server.js`, hit `/openapi.json`, fed it through `jq '.paths | keys'` to confirm all 32 routes listed, opened `/docs` in a browser to verify Redoc rendered correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ This project wraps that engine in a REST API built for agents: accessibility sna
 - **DOM Image Extraction** - list `<img>` src/alt and optionally return inline data URLs
 - **Deploy Anywhere** - Docker, Fly.io, Railway
 - **VNC Interactive Login** - log into sites visually via noVNC, export storage state for agent reuse
+- **OpenAPI Spec** - machine-readable API description at `/openapi.json`, rendered HTML viewer at `/docs`
 
 ## Optional Dependencies
 
@@ -80,6 +81,8 @@ npm start  # downloads Camoufox on first run (~300MB)
 ```
 
 Default port is `9377`. See [Environment Variables](#environment-variables) for all options.
+
+Once running, open `http://localhost:9377/docs` for interactive API docs, or fetch the raw spec at `http://localhost:9377/openapi.json`.
 
 ### Docker
 
--- a/lib/openapi-spec.js
+++ b/lib/openapi-spec.js
@@ -1,0 +1,518 @@
+import { readFileSync } from 'fs';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const pkg = JSON.parse(readFileSync(join(__dirname, '..', 'package.json'), 'utf8'));
+
+const userIdQuery = { name: 'userId', in: 'query', required: true, schema: { type: 'string' }, description: 'Session owner identifier.' };
+const userIdBody = { type: 'string', description: 'Session owner identifier.' };
+const tabIdPath = { name: 'tabId', in: 'path', required: true, schema: { type: 'string' }, description: 'Tab identifier returned by POST /tabs.' };
+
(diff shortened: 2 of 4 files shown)


skyfallsin:
Thanks for this, @mvanhorn! Great work on the spec and the drift-detection tests — that approach is exactly right.

We took your idea and built on it in #78: instead of a hand-written spec, we auto-generate it from `@openapi` JSDoc comments on each route handler via swagger-jsdoc. This means the spec can't drift from the code — the JSDoc lives right above each `app.get/post/delete()` call. We also swapped Redoc for [swagger-stripey](https://github.com/skyfallsin/swagger-stripey), a Stripe-style three-panel docs renderer.

You're credited as co-author on the commit. Closing this in favor of #78.

PR #70, opened 2026-04-19
feat: structured extract endpoint with JSON Schema + x-ref hints

mvanhorn:
## Summary

Adds `POST /tabs/:tabId/extract`. Takes a JSON Schema and returns a typed object by resolving per-property `x-ref` hints against the current tab's accessibility refs. Zero new runtime dependencies. ~75 lines of library code, 41 lines in `server.js`, 14 unit tests.

## Demo

https://github.com/mvanhorn/camofox-browser/releases/download/evidence-structured-extract/evidence-video.mp4

Structured extract API flow

## Why

Every agent that uses camofox-browser eventually writes the same code: take an accessibility snapshot, find the refs they care about, parse the `name` field out of each line by hand, coerce strings to numbers or booleans, trim whitespace. The extract endpoint collapses that boilerplate into one request with a JSON Schema.

Stagehand's killer primitive is `extract("title, price", zodSchema)` with an LLM under the hood. This PR ships the deterministic half of that: no inference, no LLM config, no external calls - just "look up these refs and coerce them to these types." An LLM-backed variant is a natural follow-up but adds a lot of surface area (BYO-LLM config, timeouts, retries, cost control), so it's not here.

## Request shape

```json
POST /tabs/:tabId/extract
{
  "userId": "demo",
  "schema": {
    "type": "object",
    "properties": {
      "title":   { "type": "string",  "x-ref": "e1" },
      "price":   { "type": "number",  "x-ref": "e3" },
      "inStock": { "type": "boolean", "x-ref": "e4" }
    },
    "required": ["title"]
  }
}
```

Response cases:

| Case | Status | Body |
|------|--------|------|
| Happy path | 200 | `{ ok: true, data: {...} }` |
| Required ref missing | 422 | `{ ok: false, error, snapshot }` |
| Schema invalid | 400 | `{ error: "top-level schema must have type: object" }` |
| No refs yet (snapshot never called) | 409 | `{ error: "call GET /tabs/:tabId/snapshot first", snapshot: null }` |
| userId or schema missing | 400 | `{ error: "..." }` |

## Scope notes

- `x-ref` is the only hint. It points to a ref from the last snapshot. Coercion handles `string`, `number`, `integer`, `boolean`; the value comes from the ref's `name` field (which is what the snapshot already exposes).
- `array` and nested `object` are out of scope on purpose. The goal is the minimum useful primitive.
- `ajv` was in the original plan but dropped. Inline validation keeps the dependency tree unchanged.
- LLM mode was in the original plan but dropped. The dependency footprint of BYO-LLM is significant enough to deserve its own PR and its own conversation with the maintainer.

## Testing

```
Tests:       266 passed, 266 total
```

14 new tests cover:

- Schema validation (missing, non-object, unsupported types)
- Coercion (string, integer, number, boolean with "true"/"yes"/"1" and inverse)
- Required/optional handling (throw vs null)
- Multi-property extraction in a single call
- Empty refs map

## Dogfooded

Ran the server locally, created a tab on example.com, called `/snapshot` to build the ref table, then:

```
$ curl -X POST :9378/tabs/$TAB/extract -d '{
    "userId": "demo",
    "schema": {
      "type":"object",
      "properties":{"linkText":{"type":"string","x-ref":"e1"}},
      "required":["linkText"]
    }
  }'
  { "ok": true, "data": { "linkText": "Learn more" } }
```

Also verified the 400, 409, and 422 paths return the right codes and bodies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ This project wraps that engine in a REST API built for agents: accessibility sna
 - **Deploy Anywhere** - Docker, Fly.io, Railway
 - **VNC Interactive Login** - log into sites visually via noVNC, export storage state for agent reuse
 - **OpenAPI Docs** - auto-generated spec at [`/openapi.json`](http://localhost:9377/openapi.json) and interactive docs at [`/docs`](http://localhost:9377/docs)
+- **Structured Extract** - `POST /tabs/:tabId/extract` with a JSON Schema that maps properties to snapshot refs via `x-ref`
 
 ## Optional Dependencies
 
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -1626,6 +1626,182 @@
         }
       }
     },
+    "/tabs/{tabId}/extract": {
+      "post": {
+        "tags": [
+          "Content"
+        ],
+        "summary": "Structured data extraction via JSON Schema",
+        "description": "Extracts structured data from the current page using a JSON Schema whose properties\ncarry `x-ref` hints pointing at snapshot element refs (e.g. `e1`, `e2`).  \nCall `GET /tabs/{tabId}/snapshot` first to populate the ref table.\n",
+        "parameters": [
+          {
+            "name": "tabId",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "required": [
+                  "userId",
+                  "schema"
+                ],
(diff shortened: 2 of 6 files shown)


skyfallsin:
Thanks @mvanhorn — clean implementation, merged! We added a couple things on top:

- `@openapi` JSDoc annotation for the extract route (all 6 response codes) — the project auto-generates its OpenAPI spec from these
- 15 handler-level unit tests covering the full status code matrix (200/400/404/409/422/500)
- Regenerated the OpenAPI spec

mvanhorn:
And thanks for landing this one too.

PR #586, opened 2026-04-27
feat: add CAMOFOX_ACCESS_KEY for global bearer auth on all routes

payne0420:
## Summary

Adds an optional `CAMOFOX_ACCESS_KEY` env var that, when set, requires `Authorization: Bearer <key>` on every route except `/health`, the cookie endpoint, and `/stop`. Each existing key keeps its own surface; this just fills the gap for everything else.

## Why

Right now the project ships two scoped keys:

| Env var | Routes it gates |
|---|---|
| `CAMOFOX_API_KEY` | `POST /sessions/:userId/cookies` |
| `CAMOFOX_ADMIN_KEY` | `POST /stop` |

Every other route — `/tabs`, `/navigate`, `/snapshot`, `/click`, `/type`, `/screenshot`, `/scroll`, `/links`, and especially **`POST /tabs/:tabId/evaluate`** (arbitrary JS in the browser) — is unauthenticated. The README's deployment instructions (`docker run -p 9377:9377`, Fly.io, Railway) all bind to `0.0.0.0`. In practice that means:

- Anyone who can reach the host:port can drive the browser, snapshot pages, run JS via `/evaluate`, and read whatever cookie state has been imported via the existing cookie endpoint.
- The only safe deployment today is loopback-only + SSH tunnel, or a reverse proxy with auth in front.

This PR adds a single opt-in switch that lets operators safely expose the server to a LAN, a private VPN, a Tailscale net, or even the public internet behind TLS — without forking, without a sidecar proxy, and without changing the existing key semantics.

## Design

- **Three keys, three doors.** `CAMOFOX_API_KEY` and `CAMOFOX_ADMIN_KEY` keep their existing exclusive scopes. The new `CAMOFOX_ACCESS_KEY` covers "everything else." Each can be set, rotated, or omitted independently.
- **Backwards compatible.** When `CAMOFOX_ACCESS_KEY` is unset (the default), behavior is unchanged — no new 401s, no breaking changes for anyone running today.
- **Implementation footprint is small.** One Express middleware in `server.js` (~17 lines), one config field in `lib/config.js`, one row in the README env-var table.
- **Scanner-safe.** The middleware reads only `CONFIG.accessKey` — no new `process.env` access in `server.js`, so the OpenClaw skill-scanner isolation rules in `AGENTS.md` (`env-harvesting`, `dangerous-exec`) remain satisfied.
- **Path exemptions are explicit.** `/health` (Docker healthcheck), `POST /sessions/:userId/cookies` (own key), and `POST /stop` (own key) bypass the access-key check so each key remains the sole arbiter of its surface.

## Usage

```bash
export CAMOFOX_ACCESS_KEY=$(openssl rand -hex 32)
npm start
# or via Docker / compose / Fly.io secrets

curl -H "Authorization: Bearer $CAMOFOX_ACCESS_KEY" \
  -X POST http://your-host:9377/tabs \
  -H 'Content-Type: application/json' \
  -d '{"userId":"u","sessionKey":"s","url":"https://example.com"}'
```

Without the header (or with the wrong key): `401 {"error":"Unauthorized"}`.
`/health` stays public so the Docker healthcheck (`fetch('http://localhost:9377/health')`) keeps working.

## Tests

`tests/unit/accessKey.test.js` covers:

- `CAMOFOX_ACCESS_KEY` unset → all routes open (backwards-compat)
- key set, no header → 401
- key set, wrong bearer → 401
- key set → `/health` exempt (200)
- key set + access-key bearer presented to cookie endpoint → not 401 (the cookie endpoint's own auth decides; the access middleware does not grant access there)

Pattern matches the existing `tests/unit/cookies.test.js` (custom env via `launchServer`, sequential start/stop, fetch assertions).

## Verified locally

```text
=== Test 1: /tabs WITHOUT key (expect 401) ===   → HTTP 401 {"error":"Unauthorized"}
=== Test 2: /tabs WITH wrong key (expect 401) === → HTTP 401 {"error":"Unauthorized"}
=== Test 3: /tabs WITH correct key (expect 200) === → HTTP 200 {"tabId":"..."}
=== Test 4: /health WITHOUT key (expect 200, exempt) === → HTTP 200
```

## Test plan
- [x] Unit tests added (`tests/unit/accessKey.test.js`)
- [x] Manual verification against running container (results above)
- [ ] Reviewer: `npm test -- accessKey` to run just the new suite
- [ ] Reviewer: confirm no regression in `tests/unit/cookies.test.js` (cookie endpoint should still be gated by `CAMOFOX_API_KEY` alone)

## Out of scope (happy to do in a follow-up if you'd prefer)

- A `lib/auth.js` refactor to share the gating policy (already considered — the access-key policy ("gate only when key set, fall through otherwise") differs intentionally from `requireAuth`'s ("gate always; loopback fallback in dev"), so I kept the middleware inline rather than overloading the helper).
- Adding the new var to a "Securing the server" section of the README beyond the env-var table.
- Adding `CAMOFOX_ACCESS_KEY` support to the OpenClaw plugin so the plugin can talk to a remotely-deployed, gated server.


skyfallsin:
Thanks for the PR — this fills a real security gap (especially `/evaluate` being wide open). Clean implementation, good use of timing-safe compare.

Found one critical issue and several hardening opportunities across two review rounds. Fixed in `53d8ebc` and `f19e286`:

**Bug fix: double-auth conflict**

Routes using `authMiddleware()` (traces, VNC `storage_state`) also read `Authorization: Bearer` — but for `CAMOFOX_API_KEY`. When both keys are set to different values, a single header can't satisfy both checks, making 4 routes unreachable. Fixed by having `requireAuth()` accept the access key as an alternative ("superkey").

**Defense-in-depth: conditional exemptions**

The `/stop` and cookie import exemptions are now conditional on their dedicated key being configured. Previously, if an operator set `CAMOFOX_ACCESS_KEY` but forgot `CAMOFOX_ADMIN_KEY`, `/stop` was exempt from the global middleware with no other auth protecting it. Now it's only exempt when `CAMOFOX_ADMIN_KEY` is set — otherwise the access key gates it.

**Other fixes:**
- `WWW-Authenticate: Bearer realm="camofox"` on 401 responses (RFC 7235 compliance)
- Whitespace trim on env var + bearer token (prevents `CAMOFOX_ACCESS_KEY=" "` from YAML/Docker misconfiguration)
- `requireAuth` reject guard now covers `accessKey`-only configs (prevents latent loopback bypass)
- Extracted inline middleware to `accessKeyMiddleware()` in `lib/auth.js` — matches project pattern, unit-testable
- Named middleware functions (`requireAuthCheck`, `accessKeyCheck`) for Express stack trace debuggability
- Replaced 5 integration tests (full server spawn, 120s timeouts) with 45 unit tests using mock req/res (~80ms total)
- Added `AccessKeyAuth` security scheme to OpenAPI spec

All 382 unit tests pass.

PR #651, opened 2026-04-28
Add Railway deployment configuration

### LINKED ISSUE #469: railway.toml is mentioned in README but explicitly ignored in .gitignore
## Description

The README.md states that a railway.toml is included for easy deployment on Railway:

> "railway.toml is included. For Fly.io or other remote CI, you'll need a Dockerfile..."

However, the file is entirely missing from the repository because it is explicitly listed in the .gitignore file.

## Impact

This discrepancy makes out-of-the-box deployment on Railway fail. When deploying directly from GitHub to Railway, the platform defaults to the included Dockerfile. However, the current Dockerfile uses local bind mounts (RUN --mount=type=bind,source=dist,target=/dist), which are not supported in Railway's cloud building environment (Railpack/Nixpacks), causing the build to crash immediately.

Xy2002:
## Summary

Adds a `railway.toml` so Railway deployments use the Dockerfile path.

## Why

Fixes #469.

The default `Dockerfile` depends on pre-downloaded local files from `dist/` via bind mounts, which works with the local Makefile flow but fails in Railway's remote build environment. `Dockerfile.ci` already downloads Camoufox and yt-dlp during the image build, so Railway should use that file instead.

This config also maps Railway's injected `PORT` into `CAMOFOX_PORT`, which is what the server reads first.

## Changes

- Add `railway.toml`
- Configure Railway to build with `Dockerfile.ci`
- Configure `/health` as the deployment health check
- Set restart policy for failed deployments

## Testing

- Built locally with `docker build -f Dockerfile.ci -t camofox-browser:railway-test .`
- Ran the container with Railway-style `PORT` mapping
- Verified `GET /health`
- Smoke-tested tab creation and snapshot against `https://example.com`

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This project wraps that engine in a REST API built for agents: accessibility sna
 - **Large Page Handling** - automatic snapshot truncation with offset-based pagination
 - **Download Capture** - capture browser downloads and fetch them via API (optional inline base64)
 - **DOM Image Extraction** - list `<img>` src/alt and optionally return inline data URLs
-- **Deploy Anywhere** - Docker, Fly.io
+- **Deploy Anywhere** - Docker, Fly.io, Railway
 - **VNC Interactive Login** - log into sites visually via noVNC, export storage state for agent reuse
 - **OpenAPI Docs** - auto-generated spec at [`/openapi.json`](http://localhost:9377/openapi.json) and interactive docs at [`/docs`](http://localhost:9377/docs)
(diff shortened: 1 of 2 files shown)


skyfallsin:
thanks for this @Xy2002 — clean PR, config checks out against Railway docs. i pushed a commit on top to add Railway to the README (deploy section, cookie import, env var table). merged and released in v1.8.1 🚀

PR #1587, opened 2026-04-30
feat: POST /tabs/:tabId/viewport — physically resize page via Playwright

cunninghambe:
## Summary

Adds a new endpoint that wraps Playwright's `page.setViewportSize({width, height})` so clients can trigger a real layout reflow.

The current workaround for tools that need responsive testing — clients calling `evaluate(window.resizeTo(...))` — is a no-op on non-popup windows in most modern browsers. This route is the correct primitive for headless test runners and visual-regression tools that capture each page at multiple viewports (e.g. mobile / tablet / desktop).

## Behavior

- Method: `POST /tabs/:tabId/viewport`
- Body: `{userId, width, height}` — width/height are CSS pixels in [100, 4000]
- Returns `{ok: true, width, height}` on success, 400 when out of range, 404 on unknown tab
- 150ms settle after resize so SPAs that re-render on viewport change (media queries, ResizeObserver subscribers, useMediaQuery hooks) have a chance to settle before the response

Mirrors the existing `/scroll` route shape: same error handling, same plugin event emit (`tab:viewport`), same OpenAPI documentation pattern (`@openapi` JSDoc above the handler).

## Use case

BugHunter's v0.17 multi-viewport responsive vision pass needs to capture each authenticated route at 375 / 768 / 1280 px to surface mobile-only bugs (table no horizontal-scroll, modals fullscreen on phone, sidebar-collapse failing). The matching `set_viewport` MCP tool was just added to camofox-mcp and routes to this endpoint.

## Notes

- Camoufox's stealth layer may continue to report fingerprint-resistant values for `window.innerWidth` / `screen.width` regardless of viewport — this is an orthogonal concern. The viewport set IS applied at the rendering layer (screenshots reflect the new size); JS-side overrides are a separate fingerprinting feature, not a viewport-set bug.
- Single test commit; no test suite included — happy to add a smoke test at `tests/integration/viewport.test.js` if useful, just point me at the test convention.

## Test plan

- [x] Manual smoke: `POST /tabs/<id>/viewport` with width=375, height=244 returns `{ok:true, width:375, height:244}` on a live tab
- [x] Manual smoke: `POST /tabs/<bad>/viewport` returns 404
- [x] Manual smoke: `POST /tabs/<id>/viewport` with width=50 returns 400

### DIFF
--- a/server.js
+++ b/server.js
@@ -3026,19 +3026,108 @@ app.post('/tabs/:tabId/scroll', async (req, res) => {
     const session = sessions.get(normalizeUserId(userId));
     const found = session && findTab(session, req.params.tabId);
     if (!found) return res.status(404).json({ error: 'Tab not found' });
-    
+
     const { tabState } = found;
     tabState.toolCalls++; tabState.consecutiveTimeouts = 0; tabState.consecutiveFailures = 0;
-    
+
     const isVertical = direction === 'up' || direction === 'down';
     const delta = (direction === 'up' || direction === 'left') ? -amount : amount;
     await tabState.page.mouse.wheel(isVertical ? 0 : delta, isVertical ? delta : 0);
     await tabState.page.waitForTimeout(300);
-    
+
     pluginEvents.emit('tab:scroll', { userId, tabId: req.params.tabId, direction, amount });
     res.json({ ok: true });
   } catch (err) {
     log('error', 'scroll failed', { reqId: req.reqId, error: err.message });
+    handleRouteError(err, req, res);
+  }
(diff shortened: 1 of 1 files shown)


skyfallsin:
Hey @cunninghambe — we took your approach and landed it in #2471 with you as co-author. Added full test coverage (17 unit + 4 e2e) and tightened the validation slightly (`Number.isFinite()` to catch NaN/Infinity which slip past `typeof` + range checks).

Closing this one in favor of that. Thanks for the contribution!

PR #1972, opened 2026-05-03
fix: bind server to loopback by default

MOlechowski:
## Summary
- Bind the local Camofox browser server to `127.0.0.1` by default instead of all interfaces.
- Add `CAMOFOX_HOST` / `HOST` override support for intentional exposure.
- Keep Docker deployments explicit by setting `CAMOFOX_HOST=0.0.0.0` in Dockerfiles.
- Document the new bind-host option and expose it in plugin config.

## Why
Local browser automation servers should not be reachable from VM gateway or LAN-facing interfaces by default. Loopback binding keeps local clients working while reducing accidental host-service exposure.

## Test Plan
- `node --check server.js`
- `node --check lib/config.js`
- `node --check lib/launcher.js`
- `npm test -- tests/unit/auth.test.js tests/unit/security.test.js`
- Started server with `CAMOFOX_HOST=127.0.0.1 CAMOFOX_PORT=9377` and verified listener was `127.0.0.1:9377`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,7 @@ COPY scripts/ ./scripts/
 RUN scripts/install-plugin-deps.sh
 
 ENV NODE_ENV=production
+ENV CAMOFOX_HOST=0.0.0.0
 ENV CAMOFOX_PORT=9377
 
 EXPOSE 9377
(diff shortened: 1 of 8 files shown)


MOlechowski:
Closing this PR for now.

PR #2676, opened 2026-05-08
fix: add idle tab pressure cleanup

nicha16:
## Summary
- Add `POST /pressure/cleanup` for proactive browser memory-pressure cleanup.
- Close eligible idle tabs first while preserving active, locked, queued, recently active, and minimum-per-session tabs.
- Return metadata-only cleanup summaries without URLs, titles, cookies, page content, or prompt data.

## Test Plan
- `node --check server.js`
- `git diff --check origin/master...HEAD`
- Downstream live validation: cleanup closed one eligible scratch tab while preserving at least one tab per session; forced watchdog pressure path called cleanup without closing the active sole tab.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1378,6 +1378,156 @@ function createTabState(page) {
     lastRequestedUrl: null,
     googleRetryCount: 0,
     navigateAbort: null,
+    pressureObservedAt: Date.now(),
+    pressureObservedToolCalls: 0,
+  };
+}
+
+function pressureHash(value) {
+  return crypto.createHash('sha256').update(String(value)).digest('hex').slice(0, 12);
+}
+
+function pressureLockState(tabId) {
+  const lock = tabLocks.get(tabId);
+  return {
+    active: Boolean(lock?.active),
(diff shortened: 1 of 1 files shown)


skyfallsin:
Hey, appreciate the work here — the pressure cleanup logic is well thought out. The observation-based idle tracking (requiring multiple checks before closing) is exactly the right approach, and the privacy-safe hashed response format is a nice touch.

We pulled the pressure cleanup code into v1.10.0 (with co-author credit). A couple of notes:

1. **Mouse endpoint**: The `/tabs/:tabId/mouse` endpoint is unrelated to pressure cleanup — if you'd like that merged, could you open a separate PR for it? Keeps things focused.

2. **Auto-scheduling**: We're considering adding an optional internal `setInterval` (e.g. `CAMOFOX_PRESSURE_INTERVAL_MS`) that calls cleanup on a timer so it works out of the box without external cron. If you're interested in contributing that, happy to review.

Thanks for the contribution! Closing this one since the core logic is in, but the mouse endpoint PR is welcome separately.

PR #3111, opened 2026-05-13
chore(config): make vnc plugin entry explicitly enabled

dawsman:
Tiny consistency fix. The `vnc` plugin entry in `camofox.config.json` is missing the explicit `"enabled": true` that `youtube` and `persistence` both carry.

The loader (`lib/plugins.js`) treats a missing `enabled` field as not-disabled, so this is a **no-op at runtime** — purely a docs/consistency change that makes the config self-describing.

```diff
-    "vnc": { "resolution": "1920x1080" }
+    "vnc": { "enabled": true, "resolution": "1920x1080" }
```

### DIFF
--- a/camofox.config.json
+++ b/camofox.config.json
@@ -5,6 +5,6 @@
   "plugins": {
     "youtube": { "enabled": true },
     "persistence": { "enabled": true },
-    "vnc": { "resolution": "1920x1080" }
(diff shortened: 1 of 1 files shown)


skyfallsin:
Shipped in v1.11 — thank you!

We merged the VNC config cleanup, adjusted so the plugin entry is explicit while remaining disabled by default.

skyfallsin:

PR #3377, opened 2026-05-17
feat: add coordinate mouse endpoint

nicha16:
## Summary
- Adds `POST /tabs/:tabId/mouse` for viewport CSS-pixel mouse actions (`click`, `move`, `down`, `up`).
- Validates coordinates, action, button, and click count; uses the existing per-user limit + tab lock path.
- Documents the endpoint in OpenAPI and regenerates `openapi.json`.
- Adds e2e coverage plus a test-client helper for coordinate mouse actions.

## Test Plan
- `npm run generate-openapi`
- `NODE_OPTIONS='--experimental-vm-modules' npx jest --runInBand --forceExit tests/unit/openapi.test.js`
- `NODE_OPTIONS='--experimental-vm-modules' npx jest --config jest.config.e2e.cjs --runInBand --forceExit tests/e2e/mouse.test.js`

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -1128,6 +1128,116 @@
         }
       }
     },
+    "/tabs/{tabId}/mouse": {
+      "post": {
+        "tags": [
+          "Interaction"
+        ],
+        "summary": "Send a mouse action at viewport coordinates",
+        "description": "Sends a low-level mouse action using viewport CSS-pixel coordinates. This is useful for human handoff UIs and visual automation flows where an external operator selects coordinates directly instead of an element ref or CSS selector.\n",
+        "parameters": [
+          {
+            "name": "tabId",
+            "in": "path",
+            "required": true,
+            "schema": {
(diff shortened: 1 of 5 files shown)


skyfallsin:
Thanks for the coordinate mouse endpoint. This is a useful option for screenshot or VNC handoff, but the implementation needs changes before merge.

Please rebase onto current master and:

- update `session.lastAccess` for successful mouse actions, as other interaction routes do;
- invalidate snapshot/rendered-content caches after actions that can change page state, so paginated snapshots cannot return stale pre-click content;
- strictly reject non-integer `clickCount` values instead of silently flooring them;
- regenerate OpenAPI after the rebase and fix the failing required publish check;
- add regression coverage for activity refresh and post-action snapshot freshness.

The endpoint should remain an advanced coordinate-level API; refs and selectors should remain the preferred default for normal agent interaction.

PR #4730, opened 2026-05-28
harden: bind to 127.0.0.1 by default; honor CAMOFOX_HOST

robertbentleyhq:
## Summary

`server.js` previously called `app.listen(PORT)` with no host argument, which makes Node bind to the unspecified address (`0.0.0.0` on most kernels). The launchd wrapper this server ships with on macOS exports `CAMOFOX_HOST=localhost`, but nothing in the code reads it -- so the service is reachable on every interface the host has configured. This PR closes that gap and adds documentation + tests for the new env var.

The companion global access-key gate (`CAMOFOX_ACCESS_KEY` + `accessKeyMiddleware`) is already in place and unchanged here. Binding to loopback by default gives operators a defense-in-depth layer underneath that auth so the network surface is also reduced.

### Changes

- **`lib/config.js`** — add `host` to the loaded config, defaulting to `127.0.0.1`. Trim whitespace, treat empty string as unset. Forward `CAMOFOX_HOST` through `serverEnv` so subprocesses see the same value.
- **`server.js`** — pass `HOST` to `app.listen(PORT, HOST, ...)`, log it on `server:started` and `EADDRINUSE` for operator visibility.
- **`README.md`** — document `CAMOFOX_HOST` in the env-var table with the `127.0.0.1` default and the Docker/Fly escape hatch.
- **`tests/unit/config.test.js`** — lock the loopback default, explicit-override, whitespace-trim, empty-string fallback, and `serverEnv`-forwarding contracts (5 new tests).

### Compatibility

Docker / Fly.io / Railway deployments that already want public ingress set `CAMOFOX_HOST=0.0.0.0` (or just `HOST` if their orchestrator does that) -- they continue to work unchanged. Only `npm start` on a developer laptop changes behavior: it stops listening on `0.0.0.0` and listens on loopback instead.

## Test plan

- [x] `NODE_OPTIONS='--experimental-vm-modules' npx jest tests/unit/config.test.js` -- 7 pass
- [x] `NODE_OPTIONS='--experimental-vm-modules' npx jest tests/unit/auth.test.js tests/unit/accessKey.test.js` -- 70 pass (no regressions on the existing access-key middleware)
- [x] Manual smoke: restart service, confirm `lsof -nP -iTCP:9377 -sTCP:LISTEN` shows `TCP 127.0.0.1:9377 (LISTEN)` (was `TCP *:9377` before)
- [x] Manual smoke: `curl http://127.0.0.1:9377/health` returns 200; `curl -H "Authorization: Bearer ..." http://127.0.0.1:9377/tabs?userId=test` returns 200; unauthenticated `/tabs` returns 401

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,24 +42,25 @@ RUN apt-get update && apt-get install -y \
     python3-minimal \
     && rm -rf /var/lib/apt/lists/*
 
-# Pre-bake Camoufox browser binary into image via bind mount (downloaded by Makefile)
+# Pre-bake Camoufox browser binary into image (downloaded at build time)
 # Note: unzip returns exit code 1 for warnings (Unicode filenames), so we use || true and verify
-RUN --mount=type=bind,source=dist,target=/dist \
-    mkdir -p /root/.cache/camoufox \
-    && (unzip -q /dist/camoufox-${ARCH}.zip -d /root/.cache/camoufox || true) \
+RUN mkdir -p /root/.cache/camoufox \
+    && curl -L -o /tmp/camoufox.zip "https://github.com/daijro/camoufox/releases/download/v${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}/camoufox-${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}-lin.${ARCH}.zip" \
+    && (unzip -q /tmp/camoufox.zip -d /root/.cache/camoufox || true) \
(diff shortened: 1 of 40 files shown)


skyfallsin:
Superseded by #8096, which added the configurable bind host on current master with focused configuration coverage and was released in v1.13.0. This branch now conflicts and also carries unrelated persistence/proxy changes, so we are closing it rather than merging the duplicate implementation. Thank you for raising the default-bind hardening motivation.

PR #4793, opened 2026-05-29
feat: CAMOFOX_USER_DATA_DIR env routes launch to persistent context

DaveDev42:
## Summary

Adds support for a `CAMOFOX_USER_DATA_DIR` env var. When set to a path
matching `/tmp/firefox-borrow-*`, the wrapper switches from
`firefox.launch(options)` to `firefox.launchPersistentContext(userDataDir, options)`.

When unset (default), the wrapper behaves exactly like before — same
`firefox.launch(options)` call, same ephemeral profile. Regression risk
on the default path: zero.

## Motivation

Lets an external orchestrator clone a Firefox profile into a temp dir
and have camofox-browser launch against it, so agents can act on sites
the user is already logged into (cookies, OAuth tokens, etc.) without
the wrapper handling credentials itself. We use this from a hermes-agent
skill on top of an `rsync` clone of the user's daily Firefox profile.

## Changes (3 commits)

1. **feat: env-var branch** (`62332d1`)
   - New `lib/launch-branch.js` with a pure `chooseLaunch(env)` decision
     function. Validates that the env value is absolute and starts with
     the `/tmp/firefox-borrow-` prefix (defense in depth — the consuming
     shell also validates this, but a wrapper-side check protects
     against a maliciously edited plist).
   - `server.js`'s launch block branches on `chooseLaunch(process.env)`.
     The else path is the original `firefox.launch(options)` line
     unchanged.
   - `shimNewContext(browser)` helper aliases `.newContext()` on a
     persistent context to return the context itself (Playwright forbids
     multiple contexts on a persistent launch), with a one-shot warning
     logged. Also stubs `.contexts()` to return `[self]` so downstream
     iteration works.
   - 5 unit tests in `test/launch-branch.test.js` cover the decision
     function (`node --test`; no extra dev deps).

2. **chore: clarify shimNewContext options-ignored semantics** (`0892bbd`)
   - Comment + warning message expanded to say explicitly that arguments
     passed to `newContext({viewport, permissions, …})` are dropped in
     persistent mode. Found while reviewing the shim before merge.

3. **fix: shim browser.isConnected() in persistent mode** (`8ca7001`)
   - `BrowserContext` has no `isConnected()`; the wrapper calls it at
     three sites (session getter, /health route, root route). Without
     the shim the first `/tabs` request 500s with
     `browser.isConnected is not a function` and `/health` perpetually
     reports `browserConnected:false`. Proxies through
     `browser.browser().isConnected()` when available, falls back to
     `true` (the context throws on use if the underlying browser is
     gone — optimistic is safe, the real failure surfaces at next op).
   - Found during end-to-end smoke testing on a real Camoufox instance.

## Test plan

- [x] `node --test test/launch-branch.test.js` — 5/5 pass for the
      branch decision (env unset, empty, valid path, paths outside the
      `/tmp/firefox-borrow-` prefix, relative path).
- [x] Manual smoke against real Camoufox: confirmed
      `firefox.launchPersistentContext` is invoked, agent sees prior
      cookies, env-unset path produces the same launch as before, and
      `/tabs`/`/health` work in both modes.

## Backwards compatibility

Zero regression risk on the default (env-unset) path — the original
`firefox.launch(options)` line runs unchanged.

When `CAMOFOX_USER_DATA_DIR` is set, the wrapper enters a
single-context mode. Agents requesting an isolated `newContext({…})`
get the default context back with a logged warning — sufficient for
single-site agent work, not appropriate for multi-tenant scenarios.
The warning is one-shot per browser instance to avoid log spam.

### DIFF
--- a/lib/launch-branch.js
+++ b/lib/launch-branch.js
@@ -1,0 +1,17 @@
+// Decides whether the camofox launch should be ephemeral (default
+// firefox.launch) or persistent (firefox.launchPersistentContext against
+// a cloned user profile). Pure function for testability.
+
+export function chooseLaunch(env) {
+  const raw = env.CAMOFOX_USER_DATA_DIR;
+  if (!raw) return 'launch';
+  if (!raw.startsWith('/')) {
+    throw new Error(`CAMOFOX_USER_DATA_DIR must start with /, got: ${raw}`);
+  }
+  if (!raw.startsWith('/tmp/firefox-borrow-')) {
+    throw new Error(
+      `CAMOFOX_USER_DATA_DIR must start with /tmp/firefox-borrow-, got: ${raw}`
+    );
+  }
+  return 'persistent';
+}
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "prepublishOnly": "npm run build",
     "start": "node server.js",
     "test": "NODE_OPTIONS='--experimental-vm-modules' jest --runInBand --forceExit",
+    "test:unit": "node --test test/*.test.js",
     "test:e2e": "NODE_OPTIONS='--experimental-vm-modules' jest --runInBand --forceExit tests/e2e",
     "test:plugins": "NODE_OPTIONS='--experimental-vm-modules' jest --forceExit plugins/",
     "test:live": "RUN_LIVE_TESTS=1 NODE_OPTIONS='--experimental-vm-modules' jest --runInBand --forceExit tests/live",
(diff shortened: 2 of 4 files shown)


skyfallsin:
Closing as superseded by the newer persistent-profile design in #6525, which is already labeled `considering`. We want one persistence proposal to evaluate rather than two competing launch/session models. Thank you for establishing the use case.

PR #5040, opened 2026-06-02
fix: honor BROWSER_IDLE_TIMEOUT_MS=0 as never

### LINKED ISSUE #4942: BROWSER_IDLE_TIMEOUT_MS=0 does not disable idle shutdown despite README saying 0 = never
## Bug Description

The README documents `BROWSER_IDLE_TIMEOUT_MS=0` as disabling browser idle shutdown, but the current config parsing and idle-shutdown scheduler do not support that behavior.

README currently says:

```md
| `BROWSER_IDLE_TIMEOUT_MS` | Kill browser when idle (0 = never) | `300000` (5min) |
```

However, setting `BROWSER_IDLE_TIMEOUT_MS=0` does not mean "never".

## Current Behavior

In `lib/config.js`, the value is parsed with `||` fallback:

```js
browserIdleTimeoutMs: parseInt(process.env.BROWSER_IDLE_TIMEOUT_MS) || 300000,
```

Because JavaScript treats `0` as falsy, this makes:

```bash
BROWSER_IDLE_TIMEOUT_MS=0
```

silently fall back to `300000` instead of preserving `0`.

So the documented `0 = never` behavior does not work.

There is also a second issue: even if parsing were changed to preserve `0`, the scheduler currently does not guard against zero/disabled values:

```js
function scheduleBrowserIdleShutdown() {
  if (browserIdleTimer || sessions.size > 0 || !browser) return;
  browserIdleTimer = setTimeout(async () => {
    browserIdleTimer = null;
    if (sessions.size === 0 && browser) {
      log('info', 'browser idle shutdown (no sessions)');
      await closeBrowserFully('idle_shutdown');
    }
  }, BROWSER_IDLE_TIMEOUT_MS);
}
```

If `BROWSER_IDLE_TIMEOUT_MS=0` were allowed through, `setTimeout(..., 0)` would schedule an immediate shutdown when there are no sessions, which is also not "never".

## Expected Behavior

When the user sets:

```bash
BROWSER_IDLE_TIMEOUT_MS=0
```

the browser should not be shut down by the no-sessions idle timer.

This is especially useful for VNC/noVNC workflows where the user wants the headed browser/Xvfb/x11vnc display to remain available for manual inspection even when there are temporarily no active API sessions.

## Related Timeout Variables

`SESSION_TIMEOUT_MS` and `TAB_INACTIVITY_MS` use a similar `parseInt(...) || default` pattern:

```js
sessionTimeoutMs: parseInt(process.env.SESSION_TIMEOUT_MS) || 600000,
tabInactivityMs: parseInt(process.env.TAB_INACTIVITY_MS) || 300000,
```

The README does not currently document `0 = never` for these two, but the same parsing behavior means `0` cannot be represented there either. It may be worth either:

1. explicitly documenting that `0` is not supported for those values, or
2. supporting `0` as "disabled" consistently across session/tab/browser idle reapers.

## Suggested Fix

Use explicit parsing/defaulting instead of `||`, e.g. a helper that only falls back on unset/invalid values, not on `0`:

```js
function parseIntEnv(name, fallback) {
  const raw = process.env[name];
  if (raw == null || raw === '') return fallback;
  const parsed = parseInt(raw, 10);
  return Number.isFinite(parsed) ? parsed : fallback;
}
```

Then add explicit disable guards in the relevant reapers, at minimum:

```js
function scheduleBrowserIdleShutdown() {
  if (BROWSER_IDLE_TIMEOUT_MS <= 0) return;
  if (browserIdleTimer || sessions.size > 0 || !browser) return;
  // ... existing timeout logic
}
```

If `0` should also disable session expiration and tab inactivity reaping, add similar guards around those interval checks.

## Environment

- Observed on: current `master` / `v1.11.2` local checkout
- Relevant files:
  - `README.md`
  - `lib/config.js`
  - `server.js`

mvanhorn:
## Summary
`BROWSER_IDLE_TIMEOUT_MS=0` now disables idle browser shutdown, matching the README's documented "0 = never".

## Why this matters
The README documents `BROWSER_IDLE_TIMEOUT_MS=0` as "never" (env-var table), but `lib/config.js` parsed it with `parseInt(...) || 300000`. Because `0` is falsy, an explicit `0` was silently replaced by the 5-minute default, and `scheduleBrowserIdleShutdown()` had no zero guard, so it would still arm a timer even if parsing were fixed.

`lib/config.js` now parses with `Number.isFinite`, preserving `0` while still falling back to `300000` for unset or non-numeric values, and `scheduleBrowserIdleShutdown()` returns early when the timeout is `<= 0` (logging once at startup). The default is unchanged, so existing users see no difference.

## Testing
`npm test` (unit). Added cases asserting `0` is preserved and invalid values fall back to the default, and that the scheduler arms no timer when the timeout is `0`.

Closes #4942

AI was used for assistance.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -67,6 +67,7 @@ function camoufoxExecutablePath(env = process.env) {
 
 function loadConfig() {
   const externalCamoufoxExecutable = camoufoxExecutablePath();
+  const browserIdleTimeoutMs = parseInt(process.env.BROWSER_IDLE_TIMEOUT_MS, 10);
   return {
     port: parseInt(process.env.CAMOFOX_PORT || process.env.PORT || '9377', 10),
     nodeEnv: process.env.NODE_ENV || 'development',
@@ -90,7 +91,7 @@ function loadConfig() {
     maxTabsGlobal: parseInt(process.env.MAX_TABS_GLOBAL) || 50,
     navigateTimeoutMs: parseInt(process.env.NAVIGATE_TIMEOUT_MS) || 25000,
     buildrefsTimeoutMs: parseInt(process.env.BUILDREFS_TIMEOUT_MS) || 12000,
-    browserIdleTimeoutMs: parseInt(process.env.BROWSER_IDLE_TIMEOUT_MS) || 300000,
+    browserIdleTimeoutMs: Number.isFinite(browserIdleTimeoutMs) ? browserIdleTimeoutMs : 300000,
     nativeMemRestartThresholdMb: parseInt(process.env.NATIVE_MEM_RESTART_THRESHOLD_MB) || 300,
     browserRssRestartThresholdMb: parseInt(process.env.BROWSER_RSS_RESTART_THRESHOLD_MB) || 1500,
     camoufoxExecutablePath: externalCamoufoxExecutable,
--- a/server.js
+++ b/server.js
@@ -561,7 +561,12 @@ let browserIdleTimer = null;
 let browserLaunchPromise = null;
 let browserWarmRetryTimer = null;
 
+if (BROWSER_IDLE_TIMEOUT_MS <= 0) {
+  log('info', 'browser idle shutdown disabled (BROWSER_IDLE_TIMEOUT_MS=0)');
+}
+
 function scheduleBrowserIdleShutdown() {
+  if (BROWSER_IDLE_TIMEOUT_MS <= 0) return;
   if (browserIdleTimer || sessions.size > 0 || !browser) return;
   browserIdleTimer = setTimeout(async () => {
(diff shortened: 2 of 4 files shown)


mvanhorn:
Thanks @skyfallsin for landing the BROWSER_IDLE_TIMEOUT_MS=0 handling.

mvanhorn:
Thanks @skyfallsin for taking the BROWSER_IDLE_TIMEOUT_MS=0 never-timeout fix.

mvanhorn:
Thanks @skyfallsin for landing this one - the 0 means never behavior now matches the README.

PR #5041, opened 2026-06-02
fix(vnc): honor ENABLE_VNC=1 over config enabled:false

### LINKED ISSUE #4314: VNC regression in 1.11 with Docker
Hi,
VNC was working fine in 1.10 with Docker:
```bash
docker run --rm -it -p 9377:9377 -p 6080:6080 \
  -e ENABLE_VNC=1 -e VNC_BIND=0.0.0.0 \
  ghcr.io/jo-inc/camofox-browser:1.10.0

curl -4I localhost:6080
HTTP/1.1 200 OK
```

In 1.11, it doesn't activate from environment (https://github.com/jo-inc/camofox-browser/blob/master/plugins/vnc/README.md states that environment takes precedence):
```bash

docker run --rm -it -p 9377:9377 -p 6080:6080 \
  -e ENABLE_VNC=1 -e VNC_BIND=0.0.0.0 \
  ghcr.io/jo-inc/camofox-browser:1.11.2

{"ts":"2026-05-25T13:23:43.425Z","level":"debug","msg":"plugin \"vnc\" not in camofox.config.json plugins list, skipping"}
```

With a .json configuration, it fails with:
```bash

echo '{
  "id": "camofox-browser",
  "name": "Camofox Browser",
  "plugins": {
    "vnc": { "enabled": true, "resolution": "1920x1080" }
  }
}' > /tmp/camofox.config.json


docker run --rm -it -p 9377:9377 -p 6080:6080 \
  -e ENABLE_VNC=1 -e VNC_BIND=0.0.0.0 -v /tmp/camofox.config.json:/app/camofox.config.json \
  ghcr.io/jo-inc/camofox-browser:1.11.2

[vnc-watcher] ERROR: /usr/share/novnc not found; noVNC cannot start
```

### LINKED ISSUE #4933: ENABLE_VNC=1 no longer works when vnc is disabled in camofox.config.json
## Bug Description

After `camofox.config.json` was changed to explicitly set `plugins.vnc.enabled=false`, VNC/noVNC can no longer be enabled reliably with `ENABLE_VNC=1` at runtime.

There are two related regressions:

1. The plugin loader skips the VNC plugin before the plugin's env-var override logic can run.
2. Docker builds skip installing VNC system dependencies because `scripts/install-plugin-deps.sh` also filters out plugins whose config has `enabled:false`.

This makes the documented Docker workflow with `-e ENABLE_VNC=1 -p 6080:6080` fail even though the environment variable suggests VNC should be enabled.

## Steps to Reproduce

1. Build the Docker image from current `master` / `v1.11.2` with the default `camofox.config.json`:

```bash
docker build --build-arg ARCH=aarch64 -t camofox-browser:local .
```

2. Run with VNC enabled via environment variable:

```bash
docker run -d \
  --name camofox-browser \
  -p 9377:9377 \
  -p 6080:6080 \
  -e CAMOFOX_PORT=9377 \
  -e ENABLE_VNC=1 \
  -e VNC_BIND=0.0.0.0 \
  -e VNC_RESOLUTION=1920x1080 \
  camofox-browser:local
```

3. Try to open noVNC:

```bash
curl -I http://localhost:6080/vnc.html
```

## Expected Behavior

`ENABLE_VNC=1` should enable the VNC plugin, start noVNC/websockify, and serve `http://localhost:6080/vnc.html`.

This was the behavior before `camofox.config.json` started explicitly setting:

```json
"vnc": { "enabled": false, "resolution": "1920x1080" }
```

Older config only had:

```json
"vnc": { "resolution": "1920x1080" }
```

## Actual Behavior

With current config:

```json
"vnc": { "enabled": false, "resolution": "1920x1080" }
```

`ENABLE_VNC=1` does not enable the plugin because `lib/plugins.js` filters it out first:

```js
if (pluginConf === false || (typeof pluginConf === 'object' && pluginConf.enabled === false)) continue;
```

Logs show the plugin being skipped:

```text
plugin "vnc" not in camofox.config.json plugins list, skipping
```

Additionally, the Docker build does not install VNC deps because `scripts/install-plugin-deps.sh` has similar filtering:

```js
.filter(([, v]) => v && v.enabled !== false)
```

So even if a user later bind-mounts a config with `vnc.enabled=true`, the plugin may load but fail because noVNC was never installed:

```text
[vnc-watcher] ERROR: /usr/share/novnc not found; noVNC cannot start
```

## Root Cause

The VNC plugin has its own env override:

```js
const enabled = process.env.ENABLE_VNC === '1' || pluginConfig.enabled === true;
```

But that code is never reached when `camofox.config.json` contains `enabled:false`, because the outer plugin loader uses `enabled:false` as an allow-list exclusion.

The same `enabled:false` value is also used at build time to exclude the plugin's apt dependencies, making runtime overrides impossible for Docker images built from default config.

## Possible Fixes

A few possible approaches:

1. Do not include `enabled:false` for VNC in the default `camofox.config.json`; omit `enabled` and let the plugin decide based on `ENABLE_VNC=1`.
2. Update `lib/plugins.js` so env-enabled plugins can still load even if their config has `enabled:false`.
3. Update `scripts/install-plugin-deps.sh` to install dependencies for plugins that may be enabled by env vars at runtime, especially VNC.
4. Split plugin allow-list membership from runtime enabled/disabled state, so `enabled:false` does not mean "do not load or install deps".

## Environment

- Observed on: current `master` / `v1.11.2`
- Relevant commit: `6d0e5ae chore(config): make vnc plugin entry explicitly disabled by default`
- Platform: macOS Apple Silicon building/running Docker ARM64 image

mvanhorn:
## Summary
`ENABLE_VNC=1` again enables VNC at runtime even when `camofox.config.json` sets `plugins.vnc.enabled=false`, restoring the documented Docker workflow.

VNC worked in 1.10 via `-e ENABLE_VNC=1`, and `plugins/vnc/README.md` states environment takes precedence. In 1.11, shipping `plugins.vnc.enabled=false` in the default config broke that at two layers:

- `lib/plugins.js` skipped any plugin not in the config's plugin list before its `ENABLE_VNC` override could run.
- `scripts/install-plugin-deps.sh` filtered out `enabled:false` plugins, so the Docker image shipped without noVNC/websockify (`/usr/share/novnc not found`).

A plugin can now declare an enable env var in its `plugin.json` (`plugins/vnc/plugin.json` -> `{"enableEnvVar": "ENABLE_VNC"}`); the loader honors it for config-disabled plugins. Dependency installation no longer filters on enabled state, so an env-enablable plugin keeps its system deps baked into the image. `vnc-launcher` also now passes an explicit watcher env whitelist instead of spreading `process.env`, per CONTRIBUTING.md. Default behavior (VNC off when neither config nor env enables it) is unchanged.

## Testing
`npm test` (unit). Added coverage for the loader's env-override decision, the VNC env whitelist forwarded to the server subprocess, and `resolveVncConfig` / `buildWatcherEnv`.

Closes #4933, closes #4314

AI was used for assistance.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -112,6 +112,9 @@ function loadConfig() {
       zip: process.env.PROXY_ZIP || '',
       sessionDurationMinutes: parseInt(process.env.PROXY_SESSION_DURATION_MINUTES || '10', 10),
     },
+    pluginEnv: {
+      ENABLE_VNC: process.env.ENABLE_VNC,
+    },
     // Env vars forwarded to the server subprocess
     serverEnv: {
       PATH: process.env.PATH,
@@ -141,6 +144,13 @@ function loadConfig() {
       PROXY_CITY: process.env.PROXY_CITY,
       PROXY_ZIP: process.env.PROXY_ZIP,
       PROXY_SESSION_DURATION_MINUTES: process.env.PROXY_SESSION_DURATION_MINUTES,
+      ENABLE_VNC: process.env.ENABLE_VNC,
+      VNC_RESOLUTION: process.env.VNC_RESOLUTION,
+      VNC_PASSWORD: process.env.VNC_PASSWORD,
+      VIEW_ONLY: process.env.VIEW_ONLY,
+      VNC_PORT: process.env.VNC_PORT,
+      NOVNC_PORT: process.env.NOVNC_PORT,
+      VNC_BIND: process.env.VNC_BIND,
     },
     // Crash reporter (opt-in, reports sent to Cloudflare Worker relay)
     crashReportEnabled:   process.env.CAMOFOX_CRASH_REPORT_ENABLED !== 'false',
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -67,6 +67,25 @@ const ROOT_DIR = path.join(__dirname, '..');
 const PLUGINS_DIR = path.join(ROOT_DIR, 'plugins');
 const CONFIG_PATH = path.join(ROOT_DIR, 'camofox.config.json');
 
+function envFlagEnabled(value) {
+  return ['1', 'true', 'yes', 'on'].includes(String(value || '').toLowerCase());
+}
+
+function readPluginMetadata(pluginsDir, name) {
+  try {
+    const raw = fs.readFileSync(path.join(pluginsDir, name, 'plugin.json'), 'utf-8');
+    const metadata = JSON.parse(raw);
+    return metadata && typeof metadata === 'object' ? metadata : {};
+  } catch {
+    return {};
+  }
+}
+
+function pluginEnabledByEnv(metadata, env) {
+  const { enableEnvVar } = metadata;
+  return typeof enableEnvVar === 'string' && enableEnvVar.length > 0 && envFlagEnabled(env[enableEnvVar]);
+}
+
 /**
  * Read plugin configuration from camofox.config.json.
  * Supports two formats:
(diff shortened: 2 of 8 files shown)


mvanhorn:
Appreciate the quick merge on the ENABLE_VNC override @skyfallsin.

mvanhorn:
Appreciate the merge @skyfallsin, ENABLE_VNC=1 winning over the config flag should avoid some confusing sessions.

mvanhorn:
Appreciate the quick merge @skyfallsin, ENABLE_VNC=1 winning over the config file keeps runtime overrides predictable.

PR #5078, opened 2026-06-03
fix: add CAMOFOX_DISABLE_DEFAULT_ADDONS knob to exclude default UBO addon

### LINKED ISSUE #5067: UBO addon download failure leaves broken cache and blocks browser startup after restart
## Summary
The browser engine can become permanently stuck after the default UBO addon download/extract step fails.

I am using the official Docker image on Unraid (`ghcr.io/jo-inc/camofox-browser:latest`). The API service starts, but the browser engine repeatedly fails with:

```text
manifest.json is missing. Addon path must be a path to an extracted addon.
```

In the logs, the failure is preceded by repeated download errors such as:

```text
Failed to download and extract UBO: Error: Failed to download from https://addons.mozilla.org/firefox/downloads/latest/ublock-origin/latest.xpi after 5 attempts
```

If I manually delete `/root/.cache/camoufox/addons/UBO`, the container can start again temporarily. However, after a container restart the same failure often returns.

## Environment
- Image: `ghcr.io/jo-inc/camofox-browser:latest`
- Host: Unraid 7.x
- Browser API port: `9377`
- Proxy configured via `PROXY_HOST` / `PROXY_PORT`

## Observed behavior
1. Container starts.
2. Browser warmup tries to launch Camoufox.
3. UBO download/extraction fails.
4. A broken addon directory is left behind or reused.
5. On the next launch attempt, `camoufox-js` rejects the addon path because `manifest.json` is missing.
6. `/health` reports `browserConnected: false` and `browserRunning: false`.

## Expected behavior
If the addon directory exists but does not contain a valid `manifest.json`, the server should treat it as invalid, remove it or redownload it, and retry once instead of reusing the broken path.

A server-level option to disable or exclude the default UBO addon would also help deployments where the addon download is unreliable or unwanted.

## Why this looks like an upstream cache-validation issue
The problem is not solved by the documented binary-install knobs such as:
- `CAMOUFOX_EXECUTABLE`
- `npx camoufox-js fetch`
- `npm install --ignore-scripts`
- `npm install --omit=optional`

Those cover browser binary installation and download strategy, but I could not find an official server-side setting to disable the default UBO addon or to automatically heal a broken addon cache.

## Suggested fix
1. In `maybeDownloadAddons()`, if the addon directory exists but `manifest.json` is missing, delete it and re-download.
2. In `confirmPaths()`, validate the addon directory and trigger a recovery path instead of throwing immediately.
3. Add a supported server-side option to disable default addons, or at least the default UBO addon.

## Repro notes
This is reproducible after a failed UBO download followed by a container restart. The broken state persists until the addon directory is manually removed.

mvanhorn:
## Summary

Adds a server-side `CAMOFOX_DISABLE_DEFAULT_ADDONS` env knob (`1`/`true`) that passes `exclude_addons: ['UBO']` into the existing `launchOptions()` call, so camoufox-js skips the default uBlock Origin addon download/extract entirely.

- `lib/config.js`: new `disableDefaultAddons` boolean (mirrors the `prometheusEnabled` env pattern) and forwards `CAMOFOX_DISABLE_DEFAULT_ADDONS` through the server-subprocess env whitelist.
- `server.js`: conditionally passes `exclude_addons: ['UBO']` to `launchOptions()` when the flag is set (undefined otherwise), plus a one-line info log.
- `tests/unit/config.test.js`: parsing test for set/unset/`0`/`true` plus serverEnv forwarding.
- `README.md`: documents the new env var.

## Why this matters

Issue #5067 reports that when the default UBO download from addons.mozilla.org fails, camoufox-js leaves a broken addon directory behind. On the next launch, `confirmPaths()` throws `manifest.json is missing...`, permanently wedging startup (`/health` reports `browserConnected: false`) until the user manually deletes the cache dir. The reporter explicitly asks for "a supported server-side option to disable default addons, or at least the default UBO addon" for deployments where the AMO download is unreliable or unwanted.

Setting `exclude_addons: ['UBO']` means the UBO download/extract is never attempted, so the broken-cache path can't be hit. The upstream cache-healing fix lives in the `camoufox-js` package; this PR provides the requested opt-out, which is fully in-scope here.

Default behavior is unchanged: with the flag unset, no `exclude_addons` key is passed and UBO is requested exactly as before.

## Testing

- `npx jest tests/unit/config.test.js` — 6 passed (including the new parsing + serverEnv-forwarding case).
- `node --check server.js` / `node --check lib/config.js` — syntax clean.
- `npm run build` — passes.

Fixes #5067

AI was used for assistance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,6 +615,7 @@ Reddit macros return JSON directly (no HTML parsing needed):
 | `CAMOUFOX_EXECUTABLE` | External Camoufox executable to use instead of downloading/launching the bundled cache. Must point to a Camoufox bundle with sibling resources. | - |
 | `CAMOUFOX_EXECUTABLE_PATH` | Compatibility alias for `CAMOUFOX_EXECUTABLE` | - |
 | `CAMOFOX_EXECUTABLE_PATH` | Compatibility alias for `CAMOUFOX_EXECUTABLE` | - |
+| `CAMOFOX_DISABLE_DEFAULT_ADDONS` | Set to `1`/`true` to skip downloading and launching the default uBlock Origin (UBO) addon. Useful for deployments where the addons.mozilla.org download is unreliable or unwanted (a failed download otherwise leaves a broken addon cache that blocks startup). | `0` |
 | `CAMOFOX_COOKIES_DIR` | Directory for cookie files | `~/.camofox/cookies` |
 | `CAMOFOX_PROFILE_DIR` | Directory for persisted session profiles | `~/.camofox/profiles` |
 | `CAMOFOX_TRACES_DIR` | Directory for session trace zips | `~/.camofox/traces` |
--- a/lib/config.js
+++ b/lib/config.js
@@ -97,6 +97,7 @@ function loadConfig() {
     camoufoxExecutablePath: externalCamoufoxExecutable,
     camoufoxCacheDir: camoufoxCacheDir(),
     prometheusEnabled: process.env.PROMETHEUS_ENABLED === '1' || process.env.PROMETHEUS_ENABLED === 'true',
+    disableDefaultAddons: process.env.CAMOFOX_DISABLE_DEFAULT_ADDONS === '1' || process.env.CAMOFOX_DISABLE_DEFAULT_ADDONS === 'true',
     proxy: {
       strategy: inferProxyStrategy(process.env.PROXY_STRATEGY || ''),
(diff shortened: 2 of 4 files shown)


mvanhorn:
Thanks also for the addons knob, @skyfallsin - being able to exclude the default UBO set helps automation profiles a lot.

PR #5409, opened 2026-06-08
fix: add missing await to VirtualDisplay.get() in server.js

### LINKED ISSUE #5408: fix: VirtualDisplay.get() missing await causes undefined display number on Linux
## Bug Description

In `server.js` line 955, the `VirtualDisplay.get()` call is missing `await`:

```js
// current (broken):

// should be:
```

## Root Cause

`VirtualDisplay.get()` is defined as `public async get(): Promise<string>` in `camoufox-js/src/virtdisplay.ts:97`. Without `await`, `vdDisplay` receives a Promise object instead of the display number string. 

The `sync_api.ts` in the same codebase already correctly uses `await virtualDisplay.get()` (line 47), confirming this is an oversight in the server.js path.

## Impact

On Linux systems using xvfb virtual display:
- `vdDisplay` is assigned a Promise, not a display number
- The subsequent log message shows `display: undefined` or `[object Promise]`
- The display number is not properly tracked

## Fix

One-line fix at server.js:955 — add `await` before `localVirtualDisplay.get()`.

Happy to submit a PR if this analysis is correct.

Love-JourneY:
## Fix: one word — `await`

`VirtualDisplay.get()` returns `Promise<string>`, but line 955 was missing `await`.

### Evidence
- Definition: `camoufox-js/src/virtdisplay.ts:97` → `public async get(): Promise<string>`
- Correct usage: `sync_api.ts:47` → `await virtualDisplay.get()`
- Bug: `server.js:955` → `vdDisplay = localVirtualDisplay.get()`

### Impact
On Linux with xvfb, vdDisplay received a Promise instead of a display number. Log would show undefined/null.

Closes #5408

### DIFF
--- a/server.js
+++ b/server.js
@@ -952,7 +952,7 @@ async function launchBrowserInstance() {
     try {
       if (os.platform() === 'linux') {
         localVirtualDisplay = pluginCtx.createVirtualDisplay();
-        vdDisplay = localVirtualDisplay.get();
+        vdDisplay = await localVirtualDisplay.get();
         log('info', 'xvfb virtual display started', { display: vdDisplay, attempt });
       }
     } catch (err) {


skyfallsin:

PR #5416, opened 2026-06-08
fix: detect zombie browser context in getSession() to prevent 30s tab timeouts

### LINKED ISSUE #5415: [Bug]: getSession() sync pages() check misses zombie CDP connections, causing 30s tab create timeouts
## Bug Description

Tab creation intermittently times out (30s) even when the browser process is alive and `browser.isConnected()` returns true. The issue is in `getSession()` at server.js line 1142-1148: the sync `session.context.pages()` health check passes even when the CDP connection is in a zombie state, causing subsequent `session.context.newPage()` to hang.

## Steps to Reproduce

1. Start Camofox server with pre-warmed browser
2. Wait for browser pre-warm to complete
3. Send a tab creation request (`POST /tabs`)
4. Observe: sometimes immediate success, sometimes 30s timeout

Log evidence — browser pre-warmed at 09:01:57, first tab creation timed out at 09:02:40 (43 seconds after pre-warm):

```json
{"ts":"2026-06-08T09:01:57.960Z","level":"info","msg":"browser pre-warmed","ms":11505}
{"ts":"2026-06-08T09:02:40.897Z","level":"error","msg":"tab create failed","error":"tab create timed out after 30000ms"}
```

## Expected Behavior

Tab creation should succeed within seconds, or the session context should be proactively recreated if the existing one is unresponsive.

## Actual Behavior

Tab creation hangs for 30 seconds, returns 500 error. User retries succeed because the timeout triggers internal recovery.

## Root Cause

`getSession()` at line 1142-1148 uses `session.context.pages()` as a health check — but this is synchronous and returns cached data even when the underlying CDP connection is dead. The async `newPage()` call at line 2616 then hangs waiting for CDP response.

Meanwhile, `healthState.lastSuccessfulNav` (line 633) tracks the timestamp of the last successful navigation. The active health probe (line 5949-5980) checks this every 60s — but there's a gap: between the probe interval and the next tab creation, a zombie context can go undetected.

## Proposed Fix

Add a `lastSuccessfulNav` staleness check to `getSession()` — if no successful navigation in the last 120 seconds, force context recreation before returning:

```javascript
// In getSession(), after the existing pages() check:
if (session && Date.now() - healthState.lastSuccessfulNav > 120_000) {
  log('warn', 'session context possibly stale, recreating', {
    userId: key,
    lastNavMs: Date.now() - healthState.lastSuccessfulNav
  });
  await closeSession(key, session, { reason: 'stale_context', clearDownloads: true, clearLocks: true });
  session = null;
}
```

This is lightweight (no new async ops) and leverages an existing timestamp that's already maintained by every navigation.

## Affected Component

- Camofox server.js — session management / browser lifecycle

## Environment

- Camofox Docker container (camofox-browser:135.0.1-local)
- Docker host networking
- Arch Linux host

## Debug Report

Full logs showing the pattern of pre-warm → timeout → recovery across multiple sessions available on request.

## Contribution

- [x] I'd like to fix this myself and submit a PR

Love-JourneY:
### Problem

`getSession()` at line 1142-1148 uses `session.context.pages()` as a health probe — but this is synchronous and returns cached data even when the underlying CDP connection is dead. When `newPage()` is subsequently called on a zombie context, it hangs for 30s.

### Evidence

```json
// browser pre-warmed successfully
{"ts":"2026-06-08T09:01:57.960Z","msg":"browser pre-warmed","ms":11505}

// 43 seconds later — first tab creation times out
{"ts":"2026-06-08T09:02:40.897Z","msg":"tab create failed","error":"tab create timed out after 30000ms"}

// Recovery — retry succeeds
{"ts":"2026-06-08T09:03:58.686Z","msg":"tab created","userId":"bbsoyarch"}
```

### Root Cause

```javascript
// server.js:1142-1148 — existing check
try {
    session.context.pages();  // sync — returns cached data even if CDP dead
} catch (err) {
    // only catches fully closed contexts
}
// ↓ newPage() hangs 30s on zombie context
const page = await session.context.newPage();  // line 2616
```

The project already tracks `healthState.lastSuccessfulNav` (line 633) and has an active health probe (lines 5949-5980) at 60s intervals. But between probes, a zombie context can go undetected.

### Fix

Add a staleness check after the existing `pages()` probe — if no successful navigation in 120s, recreate the context proactively:

```javascript
if (session && Date.now() - healthState.lastSuccessfulNav > 120_000) {
  log('warn', 'session context possibly stale, recreating', {
    userId: key,
    lastNavAgeMs: Date.now() - healthState.lastSuccessfulNav,
  });
  await closeSession(key, session, { reason: 'stale_context', ... });
  session = null;
}
```

### Scope

- **+11 / -0 lines**, single file: `server.js`
- No new dependencies
- Leverages existing `healthState.lastSuccessfulNav` timestamp
- Backward compatible

Closes #5415

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,6 +606,18 @@ Reddit macros return JSON directly (no HTML parsing needed):
 - `@reddit_search` - search all of Reddit, returns JSON with 25 results
 - `@reddit_subreddit` - browse a subreddit (e.g., query `"programming"` -> `/r/programming.json`)
 
+## Browser Configuration
+
+Browser behavior can be tuned in `camofox.config.json`:
+
+```json
+{
+  "newPageTimeoutMs": 10000
+}
+```
+
+`newPageTimeoutMs` controls how long tab creation waits for Firefox to create a page. If the context is unresponsive, Camofox replaces only that user's context and retries once. The default is 10 seconds.
+
 ## Environment Variables
 
 | Variable | Description | Default |
--- a/camofox.config.json
+++ b/camofox.config.json
@@ -2,6 +2,7 @@
   "id": "camofox-browser",
   "name": "Camofox Browser",
   "version": "1.6.0",
+  "newPageTimeoutMs": 10000,
   "plugins": {
     "youtube": { "enabled": true },
     "persistence": { "enabled": true },
--- a/lib/config.js
+++ b/lib/config.js
@@ -13,6 +13,17 @@ import os from 'os';
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT_DIR = join(__dirname, '..');
 
+const CONFIG_PATH = join(ROOT_DIR, 'camofox.config.json');
+
+function readCamofoxConfig(configPath = CONFIG_PATH) {
+  try {
+    const parsed = JSON.parse(readFileSync(configPath, 'utf8'));
+    return parsed && typeof parsed === 'object' ? parsed : {};
+  } catch {
+    return {};
+  }
+}
+
 /** @deprecated crashReporter config moved to Cloudflare Worker relay. */
 function readCrashReporterConfig() {
   return {};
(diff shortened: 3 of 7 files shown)


skyfallsin:
Thanks for reporting this. Your underlying diagnosis looks right: a context can become half-dead—`browser.isConnected()` and `context.pages()` still appear healthy, while the first real protocol operation, `context.newPage()`, hangs.

The global `lastSuccessfulNav` heuristic in the original implementation was risky because it wasn’t scoped to the affected context. It could destroy a healthy idle session or let a zombie session survive because unrelated browser activity refreshed the timestamp.

I’ve added a more direct recovery path to this PR:

- bound `context.newPage()` using `newPageTimeoutMs` in `camofox.config.json` (default 10 seconds);
- on timeout or a dead-context error, destroy only that user’s context through the normal session lifecycle;
- recreate it and retry page creation once within the same request;
- never restart the shared browser or use idle time as a health signal.

We also verified against the supported Camoufox/Playwright versions that closing a context promptly interrupts an in-flight `newPage()` request, so the recovery mechanism is viable.

PR #5651, opened 2026-06-11
fix: await async VirtualDisplay.get() (fixes 'cannot open display: [object Promise]' on Docker/Linux)

Kyzcreig:
## Problem

`camoufox-js`'s `VirtualDisplay.get()` is **async** (returns `Promise<string>` like `':0'`), but `launchBrowserInstance()` in `server.js` calls it **without `await`**:

```js
```

So `vdDisplay` is a `Promise`, which is then passed to `launchOptions({ virtual_display: vdDisplay })` and stringified to `"[object Promise]"`. Every Camoufox launch on Linux/Docker then fails with:

```
```

This breaks the **Docker deployment entirely** — `/health` shows `browserRunning: false`, `browserConnected: false`, the background pre-warm retries forever, and every `POST /tabs` returns `500 Internal server error`. The log shows the tell: `"xvfb virtual display started", "display": {}` (an unresolved Promise rendered as `{}`) immediately followed by the launch failure.

## Fix

Add the missing `await`:

```js
```

`launchBrowserInstance()` is already `async` and `await`s `launchOptions(...)` a few lines below, so this is safe. After the fix the log correctly shows `"display": ":0"` and `"camoufox launched"`.

## Verification

Built the image (`make build && make up`) on an arm64 Mac (Docker), confirmed:
- `GET /health` → `{"ok":true,"browserConnected":true,"browserRunning":true,...}`
- `POST /tabs` for `https://nowsecure.nl` (a Cloudflare bot-detection test page) → tab created, `GET /tabs/:id/snapshot` returns the success content (`NOWSECURE` headings) with **no Cloudflare interstitial**.

One-line change; no behavior change beyond fixing the broken Linux/virtual-display launch path.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,12 @@
 FROM node:22-slim AS camofox-browser
 
 # Pinned Camoufox version for reproducible builds
-# Update these when upgrading Camoufox
-ARG CAMOUFOX_VERSION=135.0.1
-ARG CAMOUFOX_RELEASE=beta.24
+# Update these when upgrading Camoufox. NOTE: the GitHub release-tag suffix
+# (CAMOUFOX_RELEASE, e.g. beta.25) can DIFFER from the asset-filename suffix
+# (e.g. alpha.25/alpha.26) — the Makefile handles the filename; this arg is only
+# recorded into version.json for provenance.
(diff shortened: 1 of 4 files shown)


skyfallsin:

PR #5723, opened 2026-06-13
fix(reporter): add resetNativeMemBaseline no-op when crash reporting disabled

### LINKED ISSUE #5434: Disabled crash reporter lacks resetNativeMemBaseline and throws on browser close
## Bug description

When crash/hang reporting is disabled with `CAMOFOX_CRASH_REPORT_ENABLED=false`, browser close / idle cleanup can throw an unhandled rejection:

```text
TypeError: reporter.resetNativeMemBaseline is not a function
```

The server still runs, but every browser close path can hit this because `server.js` calls `reporter.resetNativeMemBaseline()` unconditionally.

## Steps to reproduce

Using a local npm install of `@askjo/camofox-browser`:

```bash
CAMOFOX_CRASH_REPORT_ENABLED=false \
CAMOFOX_PORT=9377 \
npm start
```

Then create and close a browser tab, or wait for idle shutdown after a tab was opened:

```bash
curl -sS -X POST http://127.0.0.1:9377/tabs \
  -H 'Content-Type: application/json' \
  -d '{"userId":"repro","url":"https://example.com"}'

# Then close the returned tabId, or let the browser idle timeout run.
```

## Expected behavior

Disabling crash reporting should only disable outbound reporting. The no-op reporter should preserve the same method surface as the enabled reporter, so browser cleanup can complete without an unhandled rejection.

## Actual behavior

The service logs an unhandled rejection during browser cleanup:

```text
{"level":"error","msg":"unhandledRejection","reason":"TypeError: reporter.resetNativeMemBaseline is not a function"}
```

Observed locally twice, once during idle shutdown and once during restart while a browser instance was active.

## Root cause

`server.js` unconditionally calls:

```js
reporter.resetNativeMemBaseline();
```

But the disabled/no-op branch in `lib/reporter.js` returns only:

```js
{
  reportCrash: async () => {},
  reportHang: async () => {},
  reportStuckLoop: async () => {},
  startWatchdog: () => {},
  trackRoute: () => {},
  stop: () => {},
  _anonymize: anonymize,
  _stackSignature: stackSignature,
}
```

The enabled reporter does include `resetNativeMemBaseline`, so the interface diverges only when reporting is disabled.

## Suggested fix

Add a no-op method to the disabled reporter return object:

```js
resetNativeMemBaseline: () => {},
```

This keeps the reporter interface stable without changing reporting behavior.

## Contribution/security notes

I checked `CONTRIBUTING.md` before filing. This issue/fix should not require loading `.env`, spreading `process.env`, or changing child-process env handling. It only makes the no-op reporter match the enabled reporter's method surface.

## Environment

- Package: `@askjo/camofox-browser@1.11.2`
- Node: `v25.9.0`
- npm: `11.12.1`
- OS: Ubuntu Linux x86_64, kernel `6.8.0-124-generic`
- Install mode: local npm install + systemd user service
- Relevant env: `CAMOFOX_CRASH_REPORT_ENABLED=false`

## Local validation

Applied the suggested one-line no-op locally and verified:

```text
node --check server.js
node --check lib/reporter.js
service active
browser navigate/close smoke test succeeded
no new unhandledRejection after restart + close smoke
```

I also searched for an existing issue with these exact phrases and did not find a duplicate:

- `reporter.resetNativeMemBaseline is not a function`
- `CAMOFOX_CRASH_REPORT_ENABLED resetNativeMemBaseline`

melxusgid:
## Bug

When crash/hang reporting is disabled with `CAMOFOX_CRASH_REPORT_ENABLED=false`, browser close / idle cleanup throws:

```
TypeError: reporter.resetNativeMemBaseline is not a function
```

## Root cause

`server.js` calls `reporter.resetNativeMemBaseline()` unconditionally on browser close, but the no-op reporter (returned when `enabled === false`) only stubs `reportCrash`, `reportHang`, `reportStuckLoop`, `startWatchdog`, `trackRoute`, and `stop` — it was missing `resetNativeMemBaseline`.

## Fix

Add `resetNativeMemBaseline: () => {}` to the no-op reporter return object in `lib/reporter.js:665`. This keeps the reporter interface stable without changing any reporting behavior.

## Validation

- `node --check lib/re` passes
- `node --check server.js` passes
- One-line change, no new deps

Fixes #5434

### DIFF
--- a/lib/reporter.js
+++ b/lib/reporter.js
@@ -661,6 +661,7 @@ export function createReporter(config) {
       startWatchdog: () => {},
       trackRoute: () => {},
       stop: () => {},
+      resetNativeMemBaseline: () => {},
       _anonymize: anonymize,
       _stackSignature: stackSignature,
     };


skyfallsin:
Thanks for the fix. The disabled reporter contract is now addressed by #8106 and released in [v1.12.0](https://github.com/jo-inc/camofox-browser/releases/tag/v1.12.0). Closing as superseded.

PR #5724, opened 2026-06-13
feat(snapshot): add ref annotation bounding boxes to snapshot response

### LINKED ISSUE #5086: Feature Request: Overlay element ref IDs on screenshots
Screenshots (GET /tabs/{tabId}/screenshot) currently show the raw page image without any element reference IDs. These ref IDs (e.g. e1, e54, e73) ARE available in the accessibility snapshot but not on the image itself.

This makes human-in-the-loop workflows difficult: a user sees the screenshot but has no way to know which ref ID corresponds to which button.

Request: an optional query parameter (e.g. showRefs=true) on the screenshot endpoint that overlays each interactive element's ref ID as a numbered badge on the image. The positions could be derived from the accessibility tree bounding boxes.

melxusgid:
## Feature

Adds `annotateScreenshot=true` query parameter to `GET /tabs/{tabId}/snapshot`.

When set, the response includes a `screenshotAnnotations` array with element bounding boxes:

```json
{
  "screenshotAnnotations": [
    { "refId": "e1", "role": "button", "name": "Search", "x": 120, "y": 45, "width": 200, "height": 40 },
    { "refId": "e2", "role": "link", "name": "Sign in", "x": 800, "y": 20, "width": 80, "height": 24 }
  ]
}
```

This makes human-in-the-loop workflows possible: clients can overlay ref IDs on the screenshot image.

## Implementation

New helper `getRefAnnotations(page, refs)` resolves each ref to its bounding box using the same `refToLocator` logic already used by click/type endpoints. Skips elements that are no longer visible.

Works in all three snapshot paths:
- Cached offset chunks
- Google SERP fast path
- Full ariaSnapshot

No new dependencies. Pure Playwright boundingBox() calls.

Fixes #5086

### DIFF
--- a/server.js
+++ b/server.js
@@ -2284,6 +2284,38 @@ function refToLocator(page, ref, refs) {
   return locator;
 }
 
+/**
+ * Resolve each ref to its bounding box on the page.
+ * Returns an array of { refId, role, name, x, y, width, height } objects.
+ * Skips refs whose elements are not currently visible.
+ */
+async function getRefAnnotations(page, refs) {
+  const annotations = [];
+  if (!page || page.isClosed() || !refs || refs.size === 0) return annotations;
+
+  for (const [refId, info] of refs) {
+    try {
+      const locator = refToLocator(page, refId, refs);
+      if (!locator) continue;
+      const box = await locator.boundingBox();
+      if (box && box.width > 0 && box.height > 0) {
+        annotations.push({
+          refId,
+          role: info.role,
+          name: info.name,
+          x: Math.round(box.x),
+          y: Math.round(box.y),
+          width: Math.round(box.width),
+          height: Math.round(box.height),
+        });
+      }
+    } catch {
+      // Element may have disappeared — skip silently
+    }
+  }
+  return annotations;
+}
+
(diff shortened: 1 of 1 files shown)


skyfallsin:
thanks for this—the human-in-the-loop motivation makes sense. I’m thinking a little more about the right API boundary here, particularly annotated images versus returning geometry without making the capture path brittle. stay tuned.

skyfallsin:
Thanks for this feature work. We do not want to extend the core snapshot response with screenshot-ref annotation rendering. Please move this into a plugin or a separate plugin-owned endpoint, so the annotation implementation and its screenshot contract can evolve without coupling to the core snapshot/reference pipeline.

For a reworked version, please also:

- use an explicit, request-budget-aware timeout for every bounding-box lookup; current core code does not permit unbounded `locator.boundingBox()` calls;
- avoid serial unbounded work across a large reference table;
- document the plugin endpoint contract and add focused coverage for ref-to-box alignment, hidden elements, iframe refs, and timeout behavior.

Please re-open or update this PR with the plugin design rather than adding the feature to the core snapshot route.

PR #6309, opened 2026-06-23
fix: await VirtualDisplay.get() — missing await crashes browser in Docker

pushtov:
`VirtualDisplay.get()` is async and returns a Promise, but the call on line 955 was missing `await`. The unresolved Promise was passed to camoufox as the display argument:

```
```

Browser launch fails and retries indefinitely in Docker/headless Linux.

Fixed by adding `await` so `vdDisplay` resolves to the actual display number (e.g. `:0`) before it reaches `launchOptions`.


skyfallsin:

PR #6525, opened 2026-06-25
feat: opt-in persistent browser profiles (CAMOFOX_PERSISTENT_PROFILES)

juniperbevensee:
## Problem

Sessions are created with `browser.newContext({ storageState })`. `storageState` persists **cookies + localStorage** but **not IndexedDB / Service Workers / Cache Storage**. Sites that keep their auth in IndexedDB — Telegram Web (WebK) is a clear example — therefore can't be restored from a persisted session: after a session reaps and a new context is created, the user is logged out even though `storage-state.json` was reloaded.

## Change

Opt-in persistent profiles. When `CAMOFOX_PERSISTENT_PROFILES=1`, a session launches a dedicated context via `firefox.launchPersistentContext(userDataDir, …)` instead of `browser.newContext(…)`, keyed by `sha256(userId)` under `CAMOFOX_USERDATA_DIR` (default `~/.camofox/userdata`). The full on-disk Firefox profile — IndexedDB included — then survives session reap/recreate (and, if the dir is on a volume, container recreate).

- Reuses the shared Xvfb display and the same camoufox `launchOptions` as the pooled browser, so persistent contexts get identical stealth.
- `storageState` injection is skipped for persistent sessions (the on-disk profile supersedes it).
- **Opt-in** because each persistent identity is its own browser process; with the flag off, the default shared-browser / ephemeral-context model is completely unchanged.

Two new config keys (`lib/config.js`): `persistentProfiles`, `userDataDir`.

## Verification

Built the image, ran two containers (flag on / flag off), and for each: opened a tab, wrote a key to IndexedDB, **closed the session**, recreated it for the same `userId`, and read the key back.

| Mode | IndexedDB value after session recreate |
|------|----------------------------------------|
| `CAMOFOX_PERSISTENT_PROFILES=1` | returned (survived) |
| default | `MISSING` (lost) |

## Notes / trade-offs

- Persistent contexts are one browser process per profile (heavier) — hence opt-in.
- Per-session granularity (rather than a global env flag) would be a natural follow-up if there's interest.

### DIFF
--- a/Dockerfile.simple
+++ b/Dockerfile.simple
@@ -1,0 +1,42 @@
+FROM node:22-slim
+
+ARG CAMOUFOX_VERSION=135.0.1
+ARG CAMOUFOX_RELEASE=beta.24
+ARG ARCH=aarch64
+
+RUN apt-get update && apt-get install -y \
+    libgtk-3-0 libdbus-glib-1-2 libxt6 libasound2 \
+    libx11-xcb1 libxcomposite1 libxcursor1 libxdamage1 \
+    libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 \
+    libegl1-mesa libgl1-mesa-dri libgbm1 \
+    xvfb fonts-liberation fonts-noto-color-emoji fontconfig \
+    ca-certificates curl unzip python3-minimal \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY dist/camoufox-${ARCH}.zip /tmp/camoufox.zip
+RUN mkdir -p /root/.cache/camoufox \
+    && (unzip -q /tmp/camoufox.zip -d /root/.cache/camoufox || true) \
+    && chmod -R 755 /root/.cache/camoufox \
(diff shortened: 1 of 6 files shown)


skyfallsin:
thanks for putting this together. I’m thinking a little more about the persistent-profile approach and whether it’s the right boundary for this particular use case before committing either way. stay tuned.

skyfallsin:
Thanks for the persistent-profile implementation and the clear verification work.

Since this PR was opened, the built-in persistence plugin gained opt-in IndexedDB storage-state support. This covers the stated login-persistence use case without adding one Firefox process and profile directory per user:

```json
{
  "plugins": {
    "persistence": {
      "indexedDB": true
    }
  }
}
```

The branch is now conflicting and also includes unrelated VNC, URL-validation, and Docker changes. A persistent-profile mode would also need a redesigned reset/delete contract so `DELETE /sessions/:userId/storage_state` clears the on-disk Firefox profile, not only the storage-state file.

I am closing this PR as superseded by the current persistence path. Please open a new focused proposal if full Firefox-profile persistence is still needed for a case that IndexedDB storage state cannot support.

PR #6788, opened 2026-06-29
fix: Xvfb display init and viewport compat on ARM64 Linux

### LINKED ISSUE #5916: 500 on POST /tabs: Browser.setDefaultViewport rejects viewport.isMobile
# 500 on `POST /tabs`: `Browser.setDefaultViewport` rejects `viewport.isMobile` ("not described in this scheme")

## Summary

Every `POST /tabs` request fails with a 500. The server calls `browser.newContext` / `setDefaultViewport` with a `viewport.isMobile` field, but the installed Camoufox binary's protocol does not recognize that property, so context creation fails. The browser launches and `/health` is fine, but no tab can ever be opened.

## Environment

- `@askjo/camofox-browser`: **1.11.2** (latest)
- Node.js: **v24.16.0**
- OS: **macOS** (Apple Silicon)
- Install method: `npx @askjo/camofox-browser`
- Engine: camoufox (auto-downloaded by `camoufox-js`)

## Steps to reproduce

1. `npx @askjo/camofox-browser` (server starts on :9377, `/health` returns `ok`, browser pre-warmed).
2. `curl -X POST http://localhost:9377/tabs -H 'Content-Type: application/json' -d '{"userId":"claude","sessionKey":"default","url":"https://example.com"}'`

## Expected

A tab is created and an accessibility snapshot is returned.

## Actual

HTTP 500:

```
browser.newContext: Protocol error (Browser.setDefaultViewport): ERROR: failed to call method 'Browser.setDefaultViewport' with parameters {
  "browserContextId": "e5270aa2-f6f1-4a7a-985d-5ea4b275479b",
  "viewport": {
    "viewportSize": { "width": 1280, "height": 720 },
    "deviceScaleFactor": 1,
    "isMobile": false
  }
}
```

## What I already tried (did not help)

- `npx camoufox-js fetch` to re-download the engine binary.
- Full clean reinstall: `rm -rf ~/.npm/_npx` then `npx @askjo/camofox-browser`.

The error is identical and deterministic on every `POST /tabs`.

## Likely cause

Version skew between the Playwright layer and the Camoufox binary. `package.json` pins `playwright-core: ^1.58.0` (which sends `isMobile` in `setDefaultViewport`) while `camoufox-js: >=0.10.0` floats and pulls a binary whose Juggler protocol schema does not include `viewport.isMobile`. The two are not aligned, so `setDefaultViewport` is rejected.

## Suggested fix

Pin `camoufox-js` (and the fetched binary) to a version whose protocol matches `playwright-core ^1.58.0`, or strip/guard the `isMobile` property when the binary schema does not support it.

mavolty:
## What broke

Two separate issues made Camofox unusable on ARM64 Linux servers running headless (no physical display):

### 1. Xvfb display never passed to browser

`VirtualDisplay.get()` is async — it spawns Xvfb and reads the display number from fd 3. But `launchBrowserInstance` called it without `await`, so the Promise object got passed as `virtual_display` instead of the string `:0`. The browser launched in headed mode with no display, and Xvfb sat there unused.

Log showed `"display":{}` instead of `"display":":0"`. That empty object is `JSON.stringify(Promise)`.

**Fix:** one word — `await`.

### 2. `viewport.isMobile` rejected by Camoufox v135 Juggler protocol

Playwright >=1.58 includes `isMobile: false` in `Browser.setDefaultViewport` CDP calls. Camoufox v135 (latest ARM64 binary) doesn't recognize that field in its Juggler schema, so context creation fails with:

```
```

Passing `viewport: null` in `newContext()` options skips the `setDefaultViewport` call entirely. The existing `page.setViewportSize()` calls (already in the codebase) handle viewport sizing per-page, so nothing is lost.

## Testing

Both fixes tested on Oracle Cloud Ampere A1 (ARM64, Ubuntu, Docker), Camoufox v135.0.1-beta.24:

- Tabs create and return accessibility snapshots
- Navigation works (example.com → icanhazip.com)
- Health check returns `browserConnected: true`
- No viewport regression

## Related

Closes #5916

### DIFF
--- a/server.js
+++ b/server.js
@@ -707,7 +707,7 @@ async function probeGoogleSearch(candidateBrowser) {
   let context = null;
   try {
     context = await candidateBrowser.newContext({
-      viewport: { width: 1280, height: 720 },
+      viewport: null,
       permissions: ['geolocation'],
     });
     const page = await context.newPage();
@@ -952,7 +952,7 @@ async function launchBrowserInstance() {
     try {
       if (os.platform() === 'linux') {
         localVirtualDisplay = pluginCtx.createVirtualDisplay();
-        vdDisplay = localVirtualDisplay.get();
+        vdDisplay = await localVirtualDisplay.get();
         log('info', 'xvfb virtual display started', { display: vdDisplay, attempt });
       }
     } catch (err) {
@@ -1177,7 +1177,7 @@ async function getSession(userId, { trace = false } = {}) {
       }
       const b = await ensureBrowser();
       const contextOptions = {
-        viewport: { width: 1280, height: 720 },
+        viewport: null,
         permissions: ['geolocation'],
       };
       // When geoip is active (proxy configured), camoufox auto-configures
@@ -5965,7 +5965,7 @@ setInterval(async () => {
   
   let testContext;
   try {
-    testContext = await browser.newContext();
+    testContext = await browser.newContext({ viewport: null });
     const page = await testContext.newPage();
     await page.goto('about:blank', { timeout: 5000 });
     await page.close();


gearwave00001:
Confirmed that this fixed the issue for me.  I'll close my PR #7165 , I fixed it pushed then saw the other PRs.  This fix is more simple single-file change.  I am having some difficulty getting VNC working but am successful in that after a few rounds with the agent tool

I'm using camofox-browser deployed as a container alongside hermes agent, both within a docker sbx shell on a LAN Linux server and me accessing the noVNC webpage from my macbook os tahoe chrome browser.

What works:
    - await fix on VirtualDisplay.get() — browser starts correctly, no [object Promise] error
    - viewport: null fix — tabs open successfully, no isMobile CDP rejection
    - Page rendering works (tested retail page)
    
    What needs fixing (separate from core fixes):
    - VNC watcher doesn't auto-detect Xvfb because it searches for :0 in process args, but Xvfb launches with -displayfd 3 (no display number visible in the arg list). x11vnc needs to be started manually.
    - VNC plugin disabled in camofox.config.json by default — requires ENABLE_VNC=1 env var to enable.

rbrahul:
Hi @mavolty thanks for your PR to address this issue.
Do you know when this fix will be merged?
And fix will be publicly available?

Thanks and regards,
Rahul

skyfallsin:

PR #6972, opened 2026-07-01
feat(persistence): add DELETE /sessions/:userId/cookies to reset login

PluginsKers:
## What
Adds `DELETE /sessions/:userId/cookies`, registered by the persistence plugin.
It forgets a user's persisted auth by deleting the persisted storage-state file
and clearing the live context's cookies when a session is active, so the next
session for that userId starts logged out.

## Why
There is no supported way to reset a persisted session to a logged-out state.
`POST /sessions/:userId/cookies` only imports (adds) cookies, and
`DELETE /sessions/:userId` tears the session down but the persistence plugin
checkpoints storage state on `session:destroying`, so cookies get re-persisted
and restored on the next createTab. Agents needing a forced fresh login
(expired session, account switch) had no primitive for it.

## Implementation
The route lives in the persistence plugin because that plugin already owns
profileDir resolution and the storage-state file layout, so it deletes exactly
the file the restore path reads. When a live context exists it also calls
context.clearCookies() so a later checkpoint cannot re-persist the stale login.
Auth reuses ctx.auth() like the other session endpoints. Unit tests cover the
live-session, no-session, and missing-file cases.

## Response
{ ok, userId, clearedLive, removedPersisted }

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -310,7 +310,7 @@ export function register(app, ctx) {
 | `auth` | `function` | `auth()` returns Express middleware enforcing API key / loopback |
 | `ensureBrowser` | `async function` | Launch browser if not running, return browser instance |
 | `getSession` | `async function` | `getSession(userId)` -- get or create a session |
-| `destroySession` | `function` | `destroySession(userId)` -- tear down a session |
+| `destroySession` | `async function` | `destroySession(userId, { reason })` -- tear down and await a session close |
 | `withUserLimit` | `async function` | `withUserLimit(userId, fn)` -- run `fn` within per-user concurrency limit |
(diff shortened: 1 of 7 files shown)


skyfallsin:
thanks for this—the missing reset primitive is a real problem. I think the endpoint should reset the full browser identity rather than only cookies, since sites may keep auth in localStorage or IndexedDB.

the cleanest API seems to be:

```text
GET    /sessions/:userId/storage_state
DELETE /sessions/:userId/storage_state
```

the delete path would close the live context without checkpointing it, wait for any in-flight checkpoint, then remove the persisted state. that clears cookies and the rest of the session together, while ensuring a later checkpoint can’t recreate it.

this keeps the API symmetric and avoids presenting cookie clearing as a complete logout.

PR #7163, opened 2026-07-02
Await server:shutdown listeners before closing sessions

### LINKED ISSUE #7162: Storage state fails to persist on container shutdown (race with context close)
## Summary

The persistence plugin's shutdown/SIGTERM handler tries to save
`storageState` for each session on container shutdown, but by the
time that handler runs the browser context has already been torn
down, so the save always fails with:

```
{"level":"warn","msg":"failed to persist storage state","userId":"joyjit","storageStatePath":"/data/profiles/<id>/storage-state.json","error":"browserContext.storageState: Target page, context or browser has been closed"}
```

## Current Behavior

Storage state is only flushed on two triggers today:

1. Idle-tab reap (30 min)
2. Explicit session destroy

On container shutdown/restart there's also a final-save attempt,
but the browser context is closed before that handler runs, so the
save is a guaranteed no-op that just logs a warning.

Seen 3 times in `docker logs camofox-browser`:

- 2026-06-30 17:38 (`userId: deploy-e2e`)
- 2026-06-30 17:54 (`userId: health-monitor`)
- 2026-07-02 16:34 (`userId: joyjit`, during a `docker restart camofox-browser`)

In each case the prior idle-flush happened to be recent enough that
on-disk state was still current, so no data was actually lost. But
that's incidental — if a restart happens shortly after real session
activity (login, cookie refresh) and before the next 30-minute idle
flush, that state is silently dropped. Restarts happen for upgrades,
crashes, or OOM, not just idle periods.

## Expected Behavior

On shutdown, storage state should be persisted *before* the browser
context is closed, so the shutdown-time save actually captures
current state instead of racing the teardown.

## Suggested Fix

In the persistence plugin's shutdown/SIGTERM handler, reorder so the
flush happens first:

```js
// current (approximate): context is closed, then save is attempted
await browserContext.close();
await persistStorageState(browserContext); // always fails, context is gone

// suggested: flush before teardown
await persistStorageState(browserContext);
await browserContext.close();
```

## Environment

- Deployment: Docker, `camofox-browser` container on parkcircus
- Observed via: `docker logs camofox-browser`, cross-referenced
  against a persistence runbook re-run on 2026-07-02

joyjit:
## Summary

Fixes #7162.

`gracefulShutdown()` fired `server:shutdown` with plain
`EventEmitter#emit`, which does not wait for async listeners. It then
called `closeAllSessions()` immediately after, racing the persistence
plugin's in-flight `context.storageState()` checkpoint against
`context.close()` for the same session. This intermittently fails
with:

```
{"level":"warn","msg":"failed to persist storage state","error":"browserContext.storageState: Target page, context or browser has been closed"}
```

and silently drops unsaved storage state if a restart happens shortly
after real session activity (login, cookie refresh) and before the
next idle-flush.

`closeSession()` already gets this right for `session:destroying` via
`pluginEvents.emitAsync()` (see the doc comment on `emitAsync` in
`lib/plugins.js`, and the ordering test in
`tests/unit/sessionDestroyingEvent.test.js` from #75).
`server:shutdown` just wasn't using the same helper.

## Fix

```diff
- pluginEvents.emit('server:shutdown', { signal });
+ await pluginEvents.emitAsync('server:shutdown', { signal });
```

This makes shutdown listeners (persistence checkpoint, vnc watcher
cleanup, etc.) run to completion before `closeAllSessions()` starts
tearing down contexts.

## Test plan

- Added `tests/unit/serverShutdownEvent.test.js`, modeled on the
  existing `sessionDestroyingEvent.test.js` pattern:
  - shutdown checkpoint completes before any context closes
  - same holds with multiple concurrent sessions
  - a regression guard that reproduces the pre-fix race with plain
    `emit()` and asserts the checkpoint fails, to pin the behavior
    this PR changes
- `npm run test:plugins` and the full unit suite pass (659/659); e2e
  suite requires a live browser/server harness not available in my
  environment, so I didn't run it.

### DIFF
--- a/server.js
+++ b/server.js
@@ -6001,7 +6001,7 @@ async function gracefulShutdown(signal) {
   if (shuttingDown) return;
   shuttingDown = true;
   log('info', 'shutting down', { signal });
-  pluginEvents.emit('server:shutdown', { signal });
+  await pluginEvents.emitAsync('server:shutdown', { signal });
 
   const forceTimeout = setTimeout(() => {
     log('error', 'shutdown timed out, forcing exit');
--- a/server.js
+++ b/server.js
@@ -985,6 +985,15 @@ async function launchBrowserInstance() {
         virtual_display: vdDisplay,
       });
       options.proxy = normalizePlaywrightProxy(options.proxy);
+      // Playwright's launcher defaults handleSIGTERM/SIGINT/SIGHUP to true,
+      // registering its own process signal handlers that send Browser.close
+      // straight to Firefox over its debug transport the instant a signal
+      // arrives -- independent of and racing ahead of our own gracefulShutdown()
+      // sequencing (including the persistence plugin's shutdown checkpoint).
+      // Disable them so gracefulShutdown() is the sole authority over shutdown.
+      options.handleSIGTERM = false;
+      options.handleSIGINT = false;
+      options.handleSIGHUP = false;
       await pluginEvents.emitAsync('browser:launching', { options });
 
       candidateBrowser = await firefox.launch(options);
--- a/server.js
+++ b/server.js
@@ -6001,8 +6001,11 @@ async function gracefulShutdown(signal) {
   if (shuttingDown) return;
   shuttingDown = true;
   log('info', 'shutting down', { signal });
-  await pluginEvents.emitAsync('server:shutdown', { signal });
 
+  // Arm the watchdog and stop accepting new connections before anything
+  // that can block or reject -- a hung or failing server:shutdown listener
+  // must not disable the force-exit safety net or widen the window where
+  // the server still accepts new sessions.
   const forceTimeout = setTimeout(() => {
     log('error', 'shutdown timed out, forcing exit');
     process.exit(1);
@@ -6011,6 +6014,10 @@ async function gracefulShutdown(signal) {
 
   server.close();
   stopMemoryReporter();
+
+  await pluginEvents.emitAsync('server:shutdown', { signal }).catch((err) => {
+    log('error', 'server:shutdown listener failed', { error: err.message });
+  });
 
   await closeAllSessions(`shutdown:${signal}`, {
     clearDownloads: false,
--- a/tests/unit/serverShutdownEvent.test.js
+++ b/tests/unit/serverShutdownEvent.test.js
@@ -1,0 +1,120 @@
+/**
+ * Tests for the server:shutdown lifecycle event ordering.
+ *
+ * The core invariant: server:shutdown listeners must finish running
+ * BEFORE gracefulShutdown() proceeds to close session contexts, so
+ * that a plugin's checkpoint-on-shutdown (e.g. persistence) can read
+ * context.storageState() while the context is still alive.
+ *
+ * server.js previously fired this event with plain EventEmitter#emit,
+ * which does not wait for async listeners. gracefulShutdown() then
+ * called closeAllSessions() immediately after, racing the in-flight
+ * checkpoint against context.close() -- intermittently failing with
+ * "Target page, context or browser has been closed".
+ *
+ * The fix: emit via pluginEvents.emitAsync() and await it before
+ * closing any session.
+ */
+
+import { jest } from '@jest/globals';
+import { createPluginEvents } from '../../lib/plugins.js';
+
+function makeMockContext() {
+  let closed = false;
+  return {
+    get closed() { return closed; },
+    close: jest.fn(async () => { closed = true; }),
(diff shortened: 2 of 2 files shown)


joyjit:
Follow-up commit pushed after a deeper review of the original change.

The first commit fixed the storage-state persist race by making `server:shutdown` blocking (`emitAsync` instead of fire-and-forget `emit`), so the persistence plugin's checkpoint completes before `closeAllSessions()` closes any context. That part is correct and unchanged.

But blocking that emit had three side effects I hadn't accounted for:

1. **`forceTimeout` (the 10s force-exit watchdog) was armed only *after* the awaited emit resolved.** If a `server:shutdown` listener hangs (e.g. `context.storageState()` against a wedged browser process — a realistic shutdown trigger, and nothing in the checkpoint chain has a timeout), the watchdog never gets armed and the process can hang indefinitely instead of force-exiting.
2. **A rejecting listener could abort the rest of shutdown.** `emitAsync` uses `Promise.all` (fail-fast), and `gracefulShutdown()` is invoked from the SIGTERM/SIGINT handlers with no `.catch`. A throw from any `server:shutdown` listener (e.g. vnc's unguarded `watcher.kill('SIGTERM')`) would skip `closeAllSessions()`, `closeBrowserFully()`, `sentryFlush()`, and `process.exit()` entirely, becoming a silently-logged `unhandledRejection` that never exits the process.
3. **`server.close()` ran later**, widening the window where the server still accepts new session-creation requests during the (now potentially multi-second) checkpoint loop.

This commit moves `forceTimeout` arming and `server.close()` back to running immediately (as they did before the first commit), and wraps the `server:shutdown` emit in a `.catch` so a failing listener logs and falls through instead of aborting cleanup. The checkpoint-before-close ordering is unchanged — this only fixes the side effects of blocking on it.

Also worth flagging for a future PR (not addressed here, out of scope): the persistence plugin registers its own `server:shutdown` listener that re-checkpoints every active session, which is largely redundant with the already-safe, already-awaited `session:destroying` hook that `closeSession()` triggers for every session during `closeAllSessions()`. A narrower fix would remove that redundant loop instead of making the shared `server:shutdown` event blocking for every listener — but it needs to preserve the safety-net case where `destroySession()`'s fire-and-forget path can orphan a session out of the main `sessions` map before `session:destroying` fires.

joyjit:
Important update: the fix as it stood (both prior commits) was **not sufficient**. I deployed it to a real production instance and reproduced the original bug **2 out of 2 times** even with the ordering fix live and confirmed in the running container.

## What was actually happening

Playwright's browser launcher defaults `handleSIGTERM`/`handleSIGINT`/`handleSIGHUP` to `true`, which makes it register **its own** `process.on('SIGTERM', ...)` handler at browser-launch time — completely independent of this app's own signal handling in `server.js`. On SIGTERM, that handler sends a `Browser.close` protocol message straight to Firefox over its debug transport, **outside `pluginEvents` entirely**. It races ahead of `gracefulShutdown()` regardless of how carefully the app's own async sequencing is ordered, because it isn't going through the app's code path at all.

This explains everything the earlier fix couldn't:
- The persist failure happened *before* `closeAllSessions()` ever ran (confirmed by log ordering).
- The browser's OS process was still alive when the failure happened (found and killed later by the existing force-kill fallback).
- The browser process runs in its own process group (confirmed via `ps -eo pid,ppid,pgid,sid`), ruling out the container runtime delivering SIGTERM to it directly — this is an **application-level** protocol message over an already-open connection, not an OS signal.

## The fix

Set `handleSIGTERM = false`, `handleSIGInt = false`, `handleSIGHUP = false` on the options passed to `firefox.launch()`, so `gracefulShutdown()` is the sole authority over shutdown sequencing. `closeBrowserFully()` already does an explicit `browser.close()` plus a force-kill fallback, so nothing is lost.

## Verification

Deployed to a live parkcircus instance and reproduced the exact failure conditions (create a session, restart immediately) with each fix state:

| State | Result (2 runs each) |
|---|---|
| Original code (fire-and-forget `emit`) | 2/2 failed — matches 3/3 historical production incidents |
| + `emitAsync` ordering fix only | 2/2 **still failed** |
| + this `handleSIGTERM: false` fix | 2/2 **succeeded** — `"storage state persisted"` |

Full suite still passes (659/659). No new unit test for this specific line — `firefox.launch()` isn't currently unit-testable in isolation without scaffolding disproportionate to a 3-line options change, and the production reproduction above is stronger evidence than a mock would provide.

Sorry for the churn on this PR — the original diagnosis (the `emitAsync` ordering) was real and worth fixing, but incomplete on its own.

PR #7223, opened 2026-07-03
feat(persistence): opt-in periodic storage-state checkpoint

ianks:
## Motivation

The persistence plugin currently checkpoints storage state only on cookie import, session teardown (`session:destroying`), and process shutdown (`server:shutdown`). An ungraceful crash (OOM, SIGKILL) between those points loses everything accumulated since the last checkpoint — including a login that happened seconds before the crash.

## What's added

An opt-in flag on the persistence plugin, **default-off, preserving current behavior exactly when unset**:

- `checkpointIntervalMs` / `CAMOFOX_CHECKPOINT_INTERVAL_MS` — when set to a positive number, all active sessions are additionally checkpointed on a timer via the existing `checkpoint(userId, context, 'periodic')` path, bounding crash data loss to the checkpoint interval instead of "since the last teardown". The timer is `unref()`'d and cleared on `server:shutdown`.

The flag follows the existing env-var-overrides-config pattern used for `profileDir`/`CAMOFOX_PROFILE_DIR`.

Periodic checkpoints are guarded against racing an event-driven checkpoint for the same `userId` (a `checkpointsInFlight` set), since `checkpoint()` is now called from two triggers instead of one.

This builds on the `session:destroying` event's stated purpose — its v1.7.2 release notes describe it as being introduced explicitly "for persistence checkpoints" — by adding a second checkpoint trigger to the same code path rather than a new mechanism.

Note: companion PR adding IndexedDB storage-state support touches adjacent lines in the same files (option resolution block, "persistence plugin enabled" log fields, docs); whichever lands second will need a trivial rebase.

## Testing

- Extended `plugins/persistence/plugin.test.js`: periodic checkpoint off by default, and periodic checkpoint firing for active sessions on the configured interval. Also loosened one pre-existing assertion (`toHaveBeenCalledWith`) to `expect.objectContaining` since the "plugin enabled" log now carries the new field.
- `npm test` (full suite, excluding `tests/e2e` and `tests/live` which need a real browser/network): 45 suites, 695 tests, all passing.
- No lint script/config exists in this repo, so none was run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,10 @@ By default, camofox persists each user's cookies and localStorage to `~/.camofox
 
 Override the directory with `CAMOFOX_PROFILE_DIR` or set `"profileDir"` in the persistence plugin config. To disable persistence, set `"persistence": { "enabled": false }` in `camofox.config.json`.
 
+Opt-in periodic checkpointing (default-off):
+
+- `checkpointIntervalMs` / `CAMOFOX_CHECKPOINT_INTERVAL_MS` -- periodically checkpoint all active sessions on this interval, in addition to the existing checkpoint-on-close/shutdown, so an ungraceful crash (OOM/SIGKILL) only loses state since the last periodic checkpoint.
+
 ### Session Tracing
 
 Capture a Playwright trace of every action in a session: page screenshots, DOM snapshots, network requests, and console output. Output is a single `.zip` file you can open in Playwright's built-in Trace Viewer.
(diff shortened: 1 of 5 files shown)


skyfallsin:
thanks for putting this together—the motivation is clear. we don’t think timer-based checkpointing belongs in core because it adds continuous serialization and coordination even when storage hasn’t changed.

it could work well as a separate plugin, though likely not one bundled with core. we’re considering a README list of community plugins, and this could be a good fit there. thank you again for the thoughtful contribution.

PR #7225, opened 2026-07-03
feat(persistence): include IndexedDB in persisted storage state

ianks:
## Motivation

The persistence plugin calls `context.storageState()` without `indexedDB: true`, which defaults to cookies+localStorage only. Sites that keep auth state in IndexedDB (e.g. Firebase Auth and other SSO flows) therefore never have their logins persisted — sessions on those sites silently do not survive restarts even with the plugin enabled.

## What's changed

Persisted storage state now includes IndexedDB **by default**: `{ indexedDB: true }` is threaded through `persistStorageState()` into `context.storageState()`.

Opt out with `"indexedDB": false` in the plugin config or `CAMOFOX_PERSIST_INDEXEDDB=false` (env overrides config, matching the existing `profileDir`/`CAMOFOX_PROFILE_DIR` pattern) to restore the previous cookies+localStorage-only behavior.

Playwright's `storageState({ indexedDB })` option exists since v1.51; `playwright-core` in this repo is `^1.58.0`, so no dependency bump is needed. Restore is unchanged — a persisted snapshot containing IndexedDB data is loaded through the same `contextOptions.storageState` path.

Note: companion PR #7223 (opt-in periodic checkpointing) touches adjacent lines in the same files (option resolution, "persistence plugin enabled" log fields, docs); whichever lands second will need a trivial rebase.

## Testing

- Extended `plugins/persistence/plugin.test.js`: IndexedDB included by default, `indexedDB: false` opt-out, and `CAMOFOX_PERSIST_INDEXEDDB=false` overriding plugin config. Also loosened one pre-existing assertion (`toHaveBeenCalledWith`) to `expect.objectContaining` since the "plugin enabled" log now carries the new field.
- `npm test` (full suite, excluding `tests/e2e` and `tests/live` which need a real browser/network): 45 suites, 696 tests, all passing.
- No lint script/config exists in this repo, so none was run.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -378,7 +378,7 @@ export function register(app, ctx) {
 | Event | Payload |
 |-------|---------|
 | `session:cookies:import` | `{ userId, count }` |
-| `session:storage:export` | `{ userId }` |
+| `session:storage:export` | `{ userId, storageState }` |
 
 #### Server
 | Event | Payload |
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ By default, camofox persists each user's cookies and localStorage to `~/.camofox
 
 Override the directory with `CAMOFOX_PROFILE_DIR` or set `"profileDir"` in the persistence plugin config. To disable persistence, set `"persistence": { "enabled": false }` in `camofox.config.json`.
 
-Storage state includes IndexedDB by default, so logins on sites that keep auth in IndexedDB (e.g. Firebase Auth) survive restarts. Set `CAMOFOX_PERSIST_INDEXEDDB=false` (or `"indexedDB": false` in the plugin config) to restore the previous cookies+localStorage-only behavior.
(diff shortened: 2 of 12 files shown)


skyfallsin:
thanks for putting this together — IndexedDB persistence is a useful addition, especially for sites that keep auth state there.

I made a few changes to keep the behavior explicit and the implementation narrow:

- IndexedDB capture is now opt-in with `"indexedDB": true` in the persistence plugin config.
- Removed the extra environment variable so there’s only one configuration path.
- VNC exports now respect the same setting as automatic persistence.
- The export checkpoint now persists the exact state that was already captured, rather than serializing the browser twice.
- Updated the docs to call out that IndexedDB may contain more than auth data and can make snapshots larger/slower.

I also added focused coverage for the default-off behavior, explicit opt-in, VNC exports, and export-triggered persistence. 38 focused tests are passing.

PR #7266, opened 2026-07-04
fix: await VirtualDisplay.get() and drop unsupported viewport isMobile on Linux launch

### LINKED ISSUE #6807: Linux: VirtualDisplay.get() not awaited + viewport isMobile protocol error
## Environment
- OS: Linux (Ubuntu 24.04)
- Node: v24.15.0
- Camofox: latest (via `npx @askjo/camofox-browser`)
- Playwright: 1.61.1
- Xvfb: 2:21.1.12-1ubuntu1.6

## Bug 1: `VirtualDisplay.get()` not awaited

**File:** `server.js:950`

```js
// Current (broken):

// Fix:
```

`VirtualDisplay.get()` is an `async` method (returns Promise). Without `await`, `vdDisplay` is a pending Promise. When stringified and passed to `launchOptions`, it becomes `"[object Promise]"` — causing the Camoufox binary to fail with:
```
```

Full log:
```
"xvfb virtual display started","display":{}  # Promise has no enumerable props
"Error: cannot open display: [object Promise]"
```

## Bug 2: Viewport `isMobile` protocol error

**File:** `server.js:1175` and `server.js:705`

```js
// Current:
viewport: { width: 1280, height: 720 }

// Fix:
viewport: null
```

When `viewport` is passed to `browser.newContext()`, Playwright 1.61 sends `Browser.setDefaultViewport` with `isMobile: false`. The Camoufox CDP schema does not include `isMobile`:

```
Protocol error (Browser.setDefaultViewport):
```

This prevents any tab from being created.

Setting `viewport: null` disables the default viewport and avoids the protocol error.

## Additional note

Snapshot JSON contains unescaped literal newlines (`\n` as 0x0a byte) inside JSON string values. `json.loads(raw, strict=False)` required as workaround.

## Patches

Both fixes applied with sed:
```bash
sed -i "s/vdDisplay = localVirtualDisplay.get();/vdDisplay = await localVirtualDisplay.get();/" server.js
sed -i "s/viewport: { width: 1280, height: 720 },/viewport: null,/" server.js
```

mvanhorn:
## Summary
`npx @askjo/camofox-browser` on Linux fails to launch a browser for two independent reasons: the Xvfb display string is a pending Promise, and Playwright 1.61 sends a `viewport.isMobile` field the Camoufox juggler rejects. This PR awaits `VirtualDisplay.get()` (async since camoufox-js 0.10) and stops passing a fixed context viewport, sizing the default virtual display at 1280x720 so non-VNC launches keep a usable display instead of Camoufox's 1x1 default.

## Why this matters
The reporter hit this on a fresh install (Ubuntu 24.04, Node 24):

> Error: cannot open display: [object Promise]

`server.js` called `localVirtualDisplay.get()` without `await`. camoufox-js 0.11.1 made `get()` async (the repo pins `>=0.10.0`), so the value stringified into the launch env as `[object Promise]`. Past that, no tab could be created: `newContext({viewport: {width: 1280, height: 720}})` makes playwright-core 1.61 emit `Browser.setDefaultViewport` with an `isMobile` property the juggler schema rejects (`Found property "<root>.viewport.isMobile" ... not described in this scheme`). A second user confirmed the same failure in the issue thread.

`await` on the older sync `get()` (0.8.x) is a no-op, so the change is compatible across the declared camoufox-js range, and the `/tabs/{tabId}/viewport` endpoint keeps working since it uses `page.setViewportSize` rather than context defaults.

## Testing
- New `tests/unit/launchCompat.test.js` (same source-contract idiom as `tests/unit/noSecrets.test.js`) locks the awaited `get()`, the removal of fixed context viewports, and the 1280x720 non-VNC display sizing.
- `NODE_OPTIONS='--experimental-vm-modules' npx jest tests/unit/launchCompat.test.js` passes; the unit suite shows no new failures vs master (the 4 suites failing on this branch fail identically on clean master).

Fixes #6807

AI was used for assistance.

### DIFF
--- a/server.js
+++ b/server.js
@@ -685,6 +685,21 @@ function getTotalTabCount() {
 // Virtual display for WebGL support and anti-detection.
 // Xvfb gives Firefox a real X display with GLX, enabling software-rendered WebGL
 // via Mesa llvmpipe. Without this, WebGL returns "no context" -- a massive bot signal.
+const DEFAULT_VIRTUAL_DISPLAY_RESOLUTION = '1280x720x24';
+
+class DefaultVirtualDisplay extends VirtualDisplay {
+  get xvfb_args() {
+    const args = super.xvfb_args;
+    const idx = args.indexOf('0');
+    if (idx > 0 && args[idx - 1] === '-screen') {
+      const patched = [...args];
+      patched[idx + 1] = DEFAULT_VIRTUAL_DISPLAY_RESOLUTION;
+      return patched;
+    }
+    return args;
+  }
+}
+
 let virtualDisplay = null;
 let browserLaunchProxy = null;
 let externalCamoufoxLaunch = null;
@@ -6055,7 +6070,7 @@ const pluginCtx = {
   metricsRegistry: getRegister,
   createMetric,
   /** Factory for Xvfb virtual display. Plugins can replace this to customise resolution/args. */
-  createVirtualDisplay: () => new VirtualDisplay(),
+  createVirtualDisplay: () => new DefaultVirtualDisplay(),
   /** The upstream VirtualDisplay class -- plugins can subclass it. */
   VirtualDisplay,
 };
--- a/server.js
+++ b/server.js
@@ -707,7 +707,7 @@ async function probeGoogleSearch(candidateBrowser) {
   let context = null;
   try {
     context = await candidateBrowser.newContext({
-      viewport: { width: 1280, height: 720 },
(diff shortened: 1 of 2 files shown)


danielnieto:
can we get this merged please?

mvanhorn:
Appreciate the pair of merges, @skyfallsin - awaiting VirtualDisplay.get() removes a real race on Linux.

PR #8219, opened 2026-07-20
feat(mcp): add stdio MCP server (Claude Code, Codex, agy, Cursor, opencode)

epicsagas:
## Update — addressed review feedback

@skyfallsin thanks for the detailed review. All four contract issues, the architecture suggestion, and the test-coverage ask are addressed below, each verified with a new test or a direct runtime check.

| Review point | Fix | Verified by |
|---|---|---|
| `camofox_import_cookies` sent `{ cookiesPath }`, REST wants `{ cookies: [...] }` | Now parses the Netscape file locally via `lib/cookies.js` (same helper OpenClaw uses) and sends the parsed array | `tests/unit/mcp-contracts.test.js` (cookie import contract) |
| `CAMOFOX_ACCESS_KEY` not forwarded | Every tool call now attaches `Authorization: Bearer <accessKey>` automatically when set (fixed for both MCP and OpenClaw) | `tests/unit/mcp-contracts.test.js` (authHeaders / fetchSpec) |
| 11 tool schemas/handlers copied between `plugin.ts` and `mcp/server.mjs` (drift risk) | Extracted into `lib/mcp-tool-contracts.mjs` — single source of truth both hosts import (schema, REST route, auth, response shaping) | `tests/unit/openclawManifest.test.js`, plus a runtime check loading the built `plugin.js` and confirming registered tools dispatch through the same `runTool()` |
| MCP server version hardcoded | Reads from `package.json` at startup | stdio smoke test asserts `serverInfo.version` matches |
| Architecture: independently installable MCP package, avoid SDK dep on every core install | `mcp/` is now its own npm package (`@askjo/camofox-mcp`, own `package.json`), depending on nothing but `@modelcontextprotocol/sdk` — no `camoufox-js`/`playwright-core`/`express`/browser download. `cd mcp && npm install` pulls ~24MB vs. the core package's 140MB+ (node_modules) + ~300MB (browser binary). The SDK reference on the root package stays as `optionalDependencies` (kept for the existing `npm install -g` / `npx` distribution path documented in `mcp/README.md`) | Verified by installing `mcp/` standalone and running the stdio smoke test against it |
| No focused mock-HTTP contract tests | Added `tests/unit/mcp-contracts.test.js` — 38 tests covering every tool's REST route/method/body, `CAMOFOX_ACCESS_KEY`/`CAMOFOX_API_KEY` auth headers, cookie parsing, screenshot decoding, and error handling | `npm test` |

Since `plugin.ts` and `mcp/server.mjs` now share `lib/mcp-tool-contracts.mjs`, the "1:1 compatibility" claim is structural, not just asserted — both hosts call the same `runTool()`/`adaptResponse()`, so they cannot drift independently.

---

## What

Adds a standalone [Model Context Protocol](https://modelcontextprotocol.io) server that exposes camofox-browser to **any MCP-compatible host** — verified for Claude Code, Codex CLI, Antigravity/agy, Cursor, and opencode. Full guide lives in **[`mcp/README.md`](https://github.com/jo-inc/camofox-browser/blob/master/mcp/README.md)**.

Because it speaks standard stdio MCP, all five hosts run the **same `camofox-mcp` bin** — only the config file differs (`mcp/README.md` §3 has a per-host snippet for each).

The MCP server mirrors the existing OpenClaw plugin **1:1**: same 11 tool names, identical JSON-Schema parameters, and the same REST routes — enforced by sharing `lib/mcp-tool-contracts.mjs` as a single source of truth.

## Why

OpenClaw isn't every agent's host. Claude Code, Codex, agy, Cursor, and opencode all want camofox as a plain MCP server. Today they'd have to re-implement the REST calls themselves. This PR ships that integration in-tree, keeping it in sync with the OpenClaw plugin automatically via a shared contract module.

## What's included

- **`lib/mcp-tool-contracts.mjs`** — canonical tool contracts (schema, REST route, auth, response shaping) shared by `mcp/server.mjs` and `plugin.ts`.
- **`mcp/server.mjs`** — stdio MCP server. Forwards every tool call to the REST API (default `localhost:9377`), attaching `CAMOFOX_ACCESS_KEY`/`CAMOFOX_API_KEY` as needed. Per-server `userId`/`sessionKey` give each MCP session an isolated camofox session.
- **`mcp/package.json`** — `mcp/` is now an independently installable package (`@askjo/camofox-mcp`), own dependency graph (`@modelcontextprotocol/sdk` only).
- **`plugin.ts`** — now imports the same canonical contracts for tool registration; `fetchApi` forwards `CAMOFOX_ACCESS_KEY`.
- **`mcp/README.md`** — full guide: architecture, 4 install options (bin/global/npx/standalone-adapter-only), per-host registration, 11-tool table, env vars (incl. `CAMOFOX_ACCESS_KEY`), troubleshooting, testing.
- **`tests/unit/mcp-contracts.test.js`** — 38 mock-HTTP contract tests (no live REST server needed).
- **`tests/unit/openclawManifest.test.js`** — now checks the OpenClaw manifest against the canonical tool list instead of a `plugin.js` regex.
- **`package.json`** — `@modelcontextprotocol/sdk` moved to `optionalDependencies`.
- **`scripts/test-mcp.mjs`** — unchanged: dependency-free handshake/schema smoke test.

## What's NOT touched

- **Main `README.md`** — unchanged.
- **REST API, server behavior** — purely additive.

## Verification

```
$ npm run test:mcp        # handshake + schema smoke test, 29 checks
$ npm test tests/unit/mcp-contracts.test.js tests/unit/openclawManifest.test.js  # 40 tests
```

Also verified end-to-end:
- MCP: `create_tab` → `snapshot` → `list_tabs` → `screenshot` → `close_tab` against a live REST server.
- OpenClaw: installed via `openclaw plugins install .`, confirmed `Status: loaded`, `openclaw plugins doctor` clean, and loaded `plugin.js` via a mock `PluginApi` to confirm all 11 tools register and dispatch through the shared contract (including cookie import producing `{ cookies }`, not a path, and `CAMOFOX_ACCESS_KEY` attached).
- Standalone `mcp/` install: `cd mcp && npm install` (~24MB, no `camoufox-js`/`playwright-core`/browser download) + stdio smoke test.

## Notes

- Follows `CONTRIBUTING.md`: child-process env whitelisted.
- The MCP server does not launch the browser — it forwards to the REST server, which must already be running.
- `mcp/README.md` documents 4 install options now (bin via source checkout / global / npx / standalone MCP-adapter-only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,30 @@ openclaw plugins install @askjo/camofox-browser
 
 **Tools:** `camofox_create_tab`  |  `camofox_snapshot`  |  `camofox_click`  |  `camofox_type`  |  `camofox_navigate`  |  `camofox_scroll`  |  `camofox_screenshot`  |  `camofox_close_tab`  |  `camofox_list_tabs`  |  `camofox_import_cookies`
 
+### MCP Server (Claude Code / any MCP host)
+
+A standalone [Model Context Protocol](https://modelcontextprotocol.io) server (`mcp/server.mjs`) exposes the same 11 tools to any MCP-compatible host — Claude Code, Cursor, etc. It mirrors the OpenClaw plugin 1:1 (same tool names, schemas, and REST routes), so behavior is identical regardless of which host reaches camofox.
+
+The MCP server only **forwards** calls — start the REST server first (`npm start`), then register the MCP server with your host:
+
+```bash
+# Claude Code
+claude mcp add camofox-browser \
+  --env CAMOFOX_BASE_URL=http://localhost:9377 \
+  -- node ./mcp/server.mjs
+```
+
+Verify with `/mcp` in Claude Code — you should see `camofox-browser` connected with all 11 tools.
+
+| Env var | Default | Purpose |
+|---------|---------|---------|
+| `CAMOFOX_BASE_URL` | `http://localhost:9377` | REST server URL |
+| `CAMOFOX_USER_ID` | `mcp-<random>` | Session isolation (one MCP server = one camofox session) |
+| `CAMOFOX_SESSION_KEY` | `default` | Tab partition within the user |
+| `CAMOFOX_API_KEY` | _(unset)_ | Required only for `camofox_import_cookies` |
+
+Run directly with `npm run mcp`, or use the `camofox-mcp` bin after install.
+
 ### Standalone
 
 Run from npm:
--- a/README.md
+++ b/README.md
@@ -78,30 +78,6 @@ openclaw plugins install @askjo/camofox-browser
 
 **Tools:** `camofox_create_tab`  |  `camofox_snapshot`  |  `camofox_click`  |  `camofox_type`  |  `camofox_navigate`  |  `camofox_scroll`  |  `camofox_screenshot`  |  `camofox_close_tab`  |  `camofox_list_tabs`  |  `camofox_import_cookies`
 
-### MCP Server (Claude Code / any MCP host)
-
-A standalone [Model Context Protocol](https://modelcontextprotocol.io) server (`mcp/server.mjs`) exposes the same 11 tools to any MCP-compatible host — Claude Code, Cursor, etc. It mirrors the OpenClaw plugin 1:1 (same tool names, schemas, and REST routes), so behavior is identical regardless of which host reaches camofox.
-
-The MCP server only **forwards** calls — start the REST server first (`npm start`), then register the MCP server with your host:
-
-```bash
-# Claude Code
-claude mcp add camofox-browser \
-  --env CAMOFOX_BASE_URL=http://localhost:9377 \
-  -- node ./mcp/server.mjs
-```
-
-Verify with `/mcp` in Claude Code — you should see `camofox-browser` connected with all 11 tools.
-
-| Env var | Default | Purpose |
-|---------|---------|---------|
-| `CAMOFOX_BASE_URL` | `http://localhost:9377` | REST server URL |
-| `CAMOFOX_USER_ID` | `mcp-<random>` | Session isolation (one MCP server = one camofox session) |
-| `CAMOFOX_SESSION_KEY` | `default` | Tab partition within the user |
-| `CAMOFOX_API_KEY` | _(unset)_ | Required only for `camofox_import_cookies` |
-
-Run directly with `npm run mcp`, or use the `camofox-mcp` bin after install.
-
 ### Standalone
 
 Run from npm:
--- a/lib/mcp-tool-contracts.mjs
+++ b/lib/mcp-tool-contracts.mjs
@@ -1,0 +1,491 @@
+/**
+ * Canonical tool contracts for the camofox-browser REST API.
+ *
+ * Single source of truth shared by two hosts that expose the same 11 tools:
+ *   - mcp/server.mjs   (stdio MCP server for Claude Code, Codex, agy, Cursor, opencode)
+ *   - plugin.ts        (OpenClaw plugin)
+ *
+ * Importing this module from both hosts means tool names, JSON-Schema
+ * parameters, REST routes, request bodies, auth semantics, and response
+ * shaping cannot drift — the REST server sees identical traffic regardless of
+ * which host the agent reached it through.
+ *
+ * Auth model (mirrors lib/auth.js):
+ *   - CAMOFOX_ACCESS_KEY (global superkey): when set, every route except
+ *     /health, cookie import, /auth-sessions, /stop requires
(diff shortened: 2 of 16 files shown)


skyfallsin:
Thanks for putting this together. The motivation is strong: a standard stdio MCP adapter would let Claude Code, Codex, Cursor, opencode, and other hosts use Camofox without each host reimplementing the REST API. We’re marking this as **considering** while we work through the integration shape.

The standalone stdio process is the right lifecycle model (rather than running MCP inside the browser server), but before merging we should address a few contract issues:

- `camofox_import_cookies` currently sends `{ cookiesPath }`, while the REST endpoint requires a parsed `{ cookies: [...] }` payload; the OpenClaw plugin parses the Netscape file locally first.
- Normal MCP requests do not forward `CAMOFOX_ACCESS_KEY`, so globally authenticated REST servers reject the advertised tools.
- The 11 tool schemas/handlers are copied from `plugin.ts` rather than sharing a source of truth, so they can drift; the current smoke test checks schema shape but not REST routes or payloads.
- The MCP server version is hardcoded.

Architecturally, this may fit best as an independently installable MCP package/adapter (potentially in this repository) that shares canonical tool contracts with the OpenClaw adapter. That would avoid adding the MCP SDK dependency to every core Camofox installation while preserving the correct per-host stdio process model.

Focused mock-HTTP contract tests for every tool—including cookie parsing, authentication, screenshots, request payloads, and error handling—would make the 1:1 compatibility claim verifiable.

epicsagas:
Thanks for the thorough review — addressed all four contract issues plus the architecture and test-coverage suggestions.

**1. Cookie import payload.** `camofox_import_cookies` now parses the Netscape file locally via `lib/cookies.js` (the same helper `plugin.ts` already used) and sends the parsed `{ cookies: [...] }` array — never a path. Covered in `tests/unit/mcp-contracts.test.js`, including a path-traversal rejection case.

**2. `CAMOFOX_ACCESS_KEY` forwarding.** Both hosts now attach `Authorization: Bearer <accessKey>` automatically on every tool call when the env var is set — fixed in `mcp/server.mjs` and, since the OpenClaw plugin had the same gap, in `plugin.ts`'s `fetchApi` too.

**3. Schema/handler drift.** Extracted the 11 tool schemas, REST routes, auth, and response shaping into `lib/mcp-tool-contracts.mjs`. Both `mcp/server.mjs` and `plugin.ts` import it and call the same `runTool()`/`adaptResponse()` — a copy-paste drift is no longer structurally possible. `tests/unit/openclawManifest.test.js` now asserts the OpenClaw manifest matches this canonical list (previously it regex-matched the built `plugin.js`).

**4. Hardcoded version.** Reads from `package.json` at startup.

**Architecture (independently installable package).** `mcp/` is now its own npm package, `@askjo/camofox-mcp` — own `package.json`, depends on nothing but `@modelcontextprotocol/sdk`. `cd mcp && npm install` pulls ~24MB with no `camoufox-js`/`playwright-core`/`express`/browser-binary download, vs. the core package's full install. I kept the root package's SDK reference as `optionalDependencies` rather than removing it outright, since `mcp/README.md` documents `npm install -g @askjo/camofox-browser` and `npx -y @askjo/camofox-browser` as supported install paths for the bin — removing it there would've broken those. `mcp/README.md` now documents both paths (Option D is the lightweight standalone one).

**Contract tests.** `tests/unit/mcp-contracts.test.js` — 38 mock-HTTP tests covering every tool's REST route/method/body, both auth headers, cookie parsing, screenshot decoding, and error handling, with no live server required.

Also re-verified the OpenClaw side end-to-end (not just unit tests): installed via `openclaw plugins install .`, confirmed `Status: loaded` / `doctor` clean, and loaded the built `plugin.js` with a mock `PluginApi` to confirm all 11 tools register and dispatch through the shared contract against a mocked REST server.

**Separate note (not part of this PR's scope).** While testing the OpenClaw side (`openclaw camofox status` / `openclaw camofox tabs`), I found two small pre-existing bugs in `plugin.ts`'s `registerCli` block:

- `openclaw camofox status` prints `Camoufox server: undefined` — it reads `health.status`, but `GET /health` returns `{ ok, engine, activeTabs, ... }` (no `status` field).
- `openclaw camofox tabs` throws `tabs is not iterable` — it assumes `GET /tabs` returns a bare array, but it actually returns `{ running, tabs: [...] }`.

Neither affects the actual MCP/OpenClaw tools agents use — only these two human-facing CLI commands. They predate this PR (`plugin.ts` wasn't in the original diff) and are unrelated to the contract issues above. Want these fixed here (since `plugin.ts` is already touched for the canonical-contracts refactor), or would you prefer a separate issue/PR?

PR #8273, opened 2026-07-20
feat: add /tabs/:tabId/upload endpoint for file attachment

leoneparise:
## Summary

Adds a `POST /tabs/:tabId/upload` endpoint that attaches a file to an upload control **without** going through the native OS file dialog (which isn't answerable in a headless/containerized browser).

## How it works

Two strategies are tried in order:

1. **Existing input** — if an `<input type="file">` is already present, call Playwright `setInputFiles` on it directly (works for hidden inputs, skips the OS dialog entirely).
2. **Triggered chooser/panel** — otherwise arm a `filechooser` listener, activate the trigger element (`ref` or `selector`) via keyboard (`focus` + `Enter`) with a forced click as fallback, then either:
   - `setFiles` on the resulting native file chooser, **or**
   - poll for an in-app panel `<input type=file>` that mounts a beat after activation and `setInputFiles` on it.

The panel-input path is **preferred** over the native chooser (preference order, not a race) so a control that surfaces *both* — e.g. LinkedIn's media picker, which A/B-tests a native chooser and a hidden panel input for the same button — attaches the file exactly once rather than producing a duplicate ("1 of 2") image.

## Request

```jsonc
POST /tabs/:tabId/upload
{
  "userId": "agent1",
  "path": "/data/photo.png",   // absolute container-side path, or string[]
  "ref": "e36",                // optional trigger ref
  "selector": "button.upload"  // optional trigger selector
}
```

`path` must be visible inside the container (e.g. a bind-mounted directory). The route guards with `fs.existsSync` and returns `400 file_not_found` otherwise. `ref`/`selector` are optional when an `input[type=file]` already exists.

Runs under the same per-user (`withUserLimit`) and per-tab (`withTabLock`) locks as the other interaction routes, refreshes element refs afterward, and emits a `tab:upload` plugin event.

## Testing

- Adds `tests/unit/uploadEndpoint.test.js` — request-validation logic plus source-contract assertions (route registered, container file guard, both attach strategies, locking), following the `typeKeyboardMode.test.js` / `navigationTimeout.test.js` convention.
- OpenAPI docs regenerated (`openapi.json`, +1 path); `openapi.test.js` "up to date" check passes.
- `28 passed` across the upload + openapi suites.

Purely additive — no existing routes or behavior changed.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -964,6 +964,11 @@
                   "selector": {
                     "type": "string",
                     "description": "Trigger element CSS/Playwright selector. Optional when an input[type=file] already exists."
+                  },
+                  "timeout": {
+                    "type": "integer",
+                    "default": 12000,
+                    "description": "Overall budget in ms to wait for an upload UI (in-app panel input or native file chooser) to appear after the trigger is activated. Ignored values (non-numeric or <= 0) fall back to the default."
                   }
                 }
               }
--- a/openapi.json
+++ b/openapi.json
@@ -912,6 +912,77 @@
         }
       }
     },
+    "/tabs/{tabId}/upload": {
+      "post": {
+        "tags": [
+          "Interaction"
+        ],
+        "summary": "Attach a file to an upload control",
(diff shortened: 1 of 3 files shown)


skyfallsin:
Thanks, @leoneparise — **PR #8273, “feat: add `/tabs/:tabId/upload` endpoint for file attachment,”** is now merged to `master`. It adds headless file attachment through existing upload inputs or file choosers.

Your original commits are preserved in merge commit `8e3cd9b`. We added a small security follow-up so each requested file must resolve inside `CAMOFOX_UPLOADS_DIR`; external paths, symlink escapes, and non-regular files are rejected. We also added a real HTTP/browser end-to-end test that uploads an in-root file and verifies the page receives it.

Thank you for the implementation and the careful upload-strategy handling.

PR #8294, opened 2026-07-21
fix(cookies): harden path traversal check in readCookieFile

andesyteoss:
## Summary

Harden the path-containment check in `readCookieFile` (`lib/cookies.js`) against a subtle path-traversal edge case (CWE-22). The existing check compared `path.resolve(cookiesDir, cookiesPath)` against a raw, un-normalized `cookiesDir + path.sep` prefix. If `cookiesDir` is provided in a non-canonical form (e.g. a trailing separator, embedded `..` segments, or a `./` component), the resolved path and the prefix disagree, and a crafted `cookiesPath` can escape the intended base directory.

This is defense-in-depth for the primary caller — `plugin.ts:505` already runs `resolve(envCfg.cookiesDir)` before calling in — but `importBootstrapCookies` and any future/external caller do not have that guarantee. The exported `readCookieFile` should be safe on its own contract.

- **CWE:** CWE-22 (Improper Limitation of a Pathname to a Restricted Directory)
- **File / function:** `lib/cookies.js` → `readCookieFile`
- **Severity:** Medium / defense-in-depth. Impact is bounded by the Netscape cookie parser (arbitrary non-cookie files fail to parse), and the primary call site already normalizes its base dir.

## Fix

Three small changes in `readCookieFile`:

1. Reject non-string / empty `cookiesPath` early with a clear error.
2. Reject absolute `cookiesPath` explicitly (previously only caught indirectly via the prefix check, which was brittle when the base wasn't normalized).
3. Normalize the base directory with `path.resolve(cookiesDir)` **before** building the `startsWith` prefix, so the comparison always uses canonical forms on both sides. Also allow the exact base directory to compare equal (edge case where `cookiesPath` resolves to the directory itself).

Diff is 9 additions / 2 removals in a single file. No behavior change for well-formed inputs.

## Data flow

- Attacker-controlled input: `cookiesPath` argument (agent tool parameter → server → `readCookieFile`).
- Sink: `fs.readFile(resolved, ...)` inside `readCookieFile`.
- Previous guard: `resolved.startsWith(cookiesDir + path.sep)` — fails when `cookiesDir` is non-canonical.
- New guard: `resolved.startsWith(path.resolve(cookiesDir) + path.sep)` plus explicit absolute-path and type checks.

## Proof of concept

Reproduces the pre-fix bypass. Run against the *pre-fix* `lib/cookies.js`:

```js
// poc.mjs — run from repo root: node poc.mjs
import { readCookieFile } from './lib/cookies.js';
import { mkdtempSync, writeFileSync, mkdirSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';

const root = mkdtempSync(join(tmpdir(), 'camo-'));
const cookiesDir = join(root, 'cookies');
const sibling    = join(root, 'cookies-evil');
mkdirSync(cookiesDir); mkdirSync(sibling);
// Netscape-format cookie file outside the intended dir
writeFileSync(join(sibling, 'stolen.txt'),
  '.example.com\tTRUE\t/\tFALSE\t0\tSECRET\tleaked\n');

// Non-canonical base (trailing '/./' — resolves fine, but the raw prefix
// used by the old startsWith check does not match the resolved path).
const nonCanonicalBase = cookiesDir + '/./';
try {
  const cookies = await readCookieFile({
    cookiesDir: nonCanonicalBase,
    cookiesPath: '../cookies-evil/stolen.txt',
    domainSuffix: 'example.com',
  });
  console.log('BYPASS — read:', cookies);
} catch (e) {
  console.log('blocked:', e.message);
}
```

Pre-fix: prints `BYPASS — read: [...]`. Post-fix: prints `blocked: cookiesPath must be a relative path within the cookies directory`.

## Testing

- Full test suite passes (17/17) with the patched `readCookieFile`.
- Verified the PoC above transitions from "BYPASS" to "blocked" after the fix.
- Verified well-formed inputs (a real cookies file under a canonical `cookiesDir`) still load unchanged.
- Verified the `plugin.ts` call site still works — it already passed a resolved base and a relative `cookiesPath`.

## Adversarial review

Before submitting, we tried to disprove this. The primary caller (`plugin.ts:505`) already runs `resolve(envCfg.cookiesDir)` on the env-configured base, so under the shipped agent flow the old check was safe against straightforward `/etc/passwd` or sibling-prefix attempts. We considered whether that made the finding moot — but `readCookieFile` and `importBootstrapCookies` are exported from `lib/cookies.js` and reachable from any future caller that forgets to pre-resolve. A library-level guard should not depend on caller hygiene. The fix is also purely additive and cannot break the current happy path.

We also confirmed the attacker precondition is realistic: `cookiesPath` originates from an agent tool argument, so a prompt-injection or malicious tool-call can drive it. Impact is limited (the file must parse as Netscape cookie format), which is why we characterize this as medium / defense-in-depth rather than high.

### DIFF
--- a/lib/cookies.js
+++ b/lib/cookies.js
@@ -52,8 +52,15 @@ function parseNetscapeCookieFile(text) {
  * @returns {Promise<Array<{name: string, value: string, domain: string, path: string, expires: number, httpOnly: boolean, secure: boolean}>>}
  */
 async function readCookieFile({ cookiesDir, cookiesPath, domainSuffix, maxBytes = 5 * 1024 * 1024 }) {
-  const resolved = path.resolve(cookiesDir, cookiesPath);
-  if (!resolved.startsWith(cookiesDir + path.sep)) {
+  if (typeof cookiesPath !== 'string' || cookiesPath.length === 0) {
+    throw new Error('cookiesPath must be a non-empty string');
+  }
+  if (path.isAbsolute(cookiesPath)) {
+    throw new Error('cookiesPath must be a relative path within the cookies directory');
+  }
+  const normalizedBase = path.resolve(cookiesDir) + path.sep;
+  const resolved = path.resolve(normalizedBase, cookiesPath);
+  if (resolved !== normalizedBase.slice(0, -1) && !resolved.startsWith(normalizedBase)) {
     throw new Error('cookiesPath must be a relative path within the cookies directory');
   }
 


skyfallsin:
Thanks @sebastionoss — merged! Your original path-validation commit is preserved in `master` history through #8582, with a follow-up that resolves symlinks and verifies the final cookie file remains inside `CAMOFOX_COOKIES_DIR`.

The follow-up carries co-author credit and adds coverage for traversal, absolute paths, internal symlinks, and symlinks that escape the cookie directory. Really appreciate the security report and clean initial fix.

PR #8319, opened 2026-07-21
Harden tab creation admission and session reaping

batumilove:
## Summary

This publishes the exact three-commit Camofox stack currently deployed and verified in production:

1. Preserve the deployed reporter compatibility behavior.
2. Add bounded tab-creation admission control.
3. Prevent the empty-session reaper from closing a session while `newPage()` is still resolving.

Admission defaults:

- global active creations: `4`
- per-user active creations: `2`
- pending queue: `8`

Queue overflow and admission timeout paths return machine-readable HTTP 429 responses with `Retry-After`. Capacity and pending-creation reservations are released on success, error, timeout, abort, and late settlement.

## Verification

- Admission/session-cleanup/config/OpenAPI: 52/52 passed.
- eBay client integration tests: 46/46 passed in its separate repository.
- Isolated Camoufox E2E with the real browser bundle: final full run 67/67 passed.
- Controlled six-create burst: one success, five bounded HTTP 429 responses, final active/pending/tabs all zero.
- Production canary after deployment: real navigation/snapshot passed; three concurrent creates passed; all four canary tabs and both canary sessions closed.
- 11-minute production soak: service active, zero restarts, zero tabs/sessions/failures, zero warning-or-higher journal entries.

## Runtime finding addressed

An isolated E2E run exposed a real race where the 60-second empty-session reaper could close a newly created session while `newPage()` was pending. The final commit adds an idempotent pending-creation reservation and tests that reaping remains blocked until every pending creation settles.

## Residual test behavior

The existing typing/click E2E can intermittently hit its 30-second click fallback timeout. Reverse-order A/B reproduced the exact same test identity and error class on both baseline and candidate. No click, typing, mouse fallback, or action-timeout code is changed here.

## Publication/rebase status

The exact reviewed production commit is preserved in the fork as `batumi/deployed-56b0a892` at `56b0a892e6f0c8b7a3d295b8be6af9af57feb5bf`.

This PR head rebases that production patch stack onto upstream `master` at `8b5b0959adfadadae38ecce9d7eed706ab102bf1`. Conflict resolution preserves newer upstream tab recycling and `createPageWithSessionRecovery` behavior. A focused TDD addition verifies that pending-creation reaper reservations follow both the original and replacement sessions during recovery. Opening this PR does not request or imply an upstream deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,10 +639,14 @@ Browser behavior can be tuned in `camofox.config.json`:
 | `CAMOFOX_TRACES_TTL_HOURS` | Traces older than this are swept on startup | `24` |
 | `MAX_SESSIONS` | Max concurrent browser sessions | `50` |
 | `MAX_TABS_PER_SESSION` | Max tabs per session | `10` |
+| `MAX_TABS_GLOBAL` | Max tabs globally and concurrent tab creations | `50` |
 | `SESSION_TIMEOUT_MS` | Session inactivity timeout | `1800000` (30min) |
 | `BROWSER_IDLE_TIMEOUT_MS` | Kill browser when idle (0 = never) | `300000` (5min) |
 | `HANDLER_TIMEOUT_MS` | Max time for any handler | `30000` (30s) |
 | `MAX_CONCURRENT_PER_USER` | Concurrent request cap per user | `3` |
+| `TAB_ADMISSION_MAX_ACTIVE` | Max concurrent tab creations globally | `4` |
+| `TAB_ADMISSION_MAX_ACTIVE_PER_USER` | Max concurrent tab creations per user | `2` |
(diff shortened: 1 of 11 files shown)


skyfallsin:
Thanks @batumilove — this surfaced an important lifecycle race: a session with no registered tabs is not necessarily idle while `newPage()` is still pending.

We are taking a focused follow-up for **issue #8555, “Tab reaper can close a session while its first `newPage()` is pending.”** It keeps the core reaper protection but uses a shared page-lease primitive across page-creation paths. We are intentionally separating that from admission queues and new `429` responses so existing clients keep their current overload behavior.

Your contribution is preserved with co-author credit in the follow-up commit. Thanks again for the detailed work and production validation.

PR #8398, opened 2026-07-26
fix: add CAMOFOX_MAX_BODY_SIZE env var to control body parser limit

### LINKED ISSUE #8397: evaluate endpoint 1MB limit is dead code — global 100KB parser runs first
## Problem

The `/tabs/:tabId/evaluate` endpoint declares a 1MB body limit:

```js
// server.js line 4572
app.post('/tabs/:tabId/evaluate', express.json({ limit: '1mb' }), ...)
```

But this is **dead code**. The global middleware on line 120 runs first and rejects any body larger than 100KB before the route-specific parser ever sees it:

```js
// server.js line 120 (from the initial commit, Feb 2026)
app.use(express.json({ limit: '100kb' }));
```

In Express, when two `express.json()` parsers are stacked, the first one that parses (or rejects) the body wins. Bodies >100KB get a `413 Payload Too Large` from the global parser, and the evaluate handler's 1MB override never fires.

## Reproduction

```bash
# Works (body <100KB):
curl -s -X POST http://localhost:9377/tabs/:tabId/evaluate \
  -H "Content-Type: application/json" \
  -d '{"userId":"test","expression":"'"$(python3 -c 'print("x"*90000)')"'"}'

# Fails with 413 (body >100KB):
curl -s -X POST http://localhost:9377/tabs/:tabId/evaluate \
  -H "Content-Type: application/json" \
  -d '{"userId":"test","expression":"'"$(python3 -c 'print("x"*200000)')"'"}'
```

## Impact

This makes it impossible to pass large expressions to evaluate, which is especially important when:
- Injecting base64-encoded binary data (e.g., file uploads via FormData + fetch)
- Passing large JSON payloads or page content through evaluate

The 100KB global limit caps effective file injection at ~73KB (after base64 expansion and JSON wrapper overhead).

## Suggested fix

Increase the global limit or move the route-specific parser before the global one. Minimum viable fix:

```diff
-app.use(express.json({ limit: '100kb' }));
+app.use(express.json({ limit: '10mb' }));
```

Or make the global limit configurable via `CAMOFOX_MAX_BODY_SIZE` env var.

igitur:
## Problem

The `/tabs/:tabId/evaluate` endpoint has a 1MB body limit override, but it's **dead code** — the global `express.json({ limit: '100kb' })` middleware on line 120 runs first and rejects any body larger than 100KB before the evaluate-specific parser ever fires.

This makes it impossible to pass large expressions to evaluate, which is important for:
- Injecting base64-encoded binary data (e.g., file uploads via FormData + fetch)
- Passing large JSON payloads through evaluate

## Fix

Add `CAMOFOX_MAX_BODY_SIZE` env var (default: `'100kb'`) to `lib/config.js` — the centralized config module where all other env vars live. Then use `CONFIG.maxBodySize` in both places:

- `lib/config.js`: `maxBodySize: process.env.CAMOFOX_MAX_BODY_SIZE || '100kb'`
- Global parser: `express.json({ limit: CONFIG.maxBodySize })`
- Evaluate parser: `express.json({ limit: CONFIG.maxBodySize })`

The cookies endpoint (`512kb`) and extract endpoint (`256kb`) are intentionally unchanged — they have their own specific limits.

## Usage

```bash
# Default (100KB, backward compatible)
camofox-browser

# Increase to 10MB for large evaluate payloads
CAMOFOX_MAX_BODY_SIZE=10mb camofox-browser
```

Closes #8397

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -94,6 +94,7 @@ function loadConfig({ configPath = CONFIG_PATH } = {}) {
     adminKey: process.env.CAMOFOX_ADMIN_KEY || '',
     apiKey: process.env.CAMOFOX_API_KEY || '',
     accessKey: (process.env.CAMOFOX_ACCESS_KEY || '').trim(),
+    maxBodySize: process.env.CAMOFOX_MAX_BODY_SIZE || '100kb',
     cookiesDir: process.env.CAMOFOX_COOKIES_DIR || join(os.homedir(), '.camofox', 'cookies'),
     profileDir: process.env.CAMOFOX_PROFILE_DIR || join(os.homedir(), '.camofox', 'profiles'),
     tracesDir: process.env.CAMOFOX_TRACES_DIR || join(os.homedir(), '.camofox', 'traces'),
--- a/server.js
+++ b/server.js
@@ -117,7 +117,7 @@ function log(level, msg, fields = {}) {
 }
 
 const app = express();
-app.use(express.json({ limit: '100kb' }));
+app.use(express.json({ limit: CONFIG.maxBodySize }));
 
 // Request logging + metrics middleware
 app.use((req, res, next) => {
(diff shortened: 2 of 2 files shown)


skyfallsin:
Thanks @igitur — merged, with your original commit retained in the history.

The follow-up keeps the global JSON limit at `100kb` and applies the configurable `CAMOFOX_EVALUATE_MAX_BODY_SIZE` limit only to `POST /tabs/:tabId/evaluate`, so the cookie and extract limits are not widened. It also forwards the setting to plugin-launched server subprocesses, documents it, and adds config and E2E coverage.

PR #8653, opened 2026-08-03
fix(docker): install build toolchain and copy mcp/ directory

### LINKED ISSUE #8639: Docker build fails on latest master: better-sqlite3 cannot build because make is missing
## Description

Building the latest `master` branch fails during the `camofox-browser` Docker image build.

The failure occurs at:

```dockerfile
RUN npm ci --omit=dev
```

`better-sqlite3` attempts to compile its native module using `node-gyp`, but the container does not have `make` installed.

The `v1.13.0` release builds and runs successfully in the same environment.

## Steps to reproduce

```bash
git clone https://github.com/jo-inc/camofox-browser.git
cd camofox-browser
make reset
```

## Expected behavior

The Docker image should build successfully, as it does with the `v1.13.0` release.

## Actual behavior

The build fails with the following error:

```text
=> ERROR [camofox-browser  8/14] RUN npm ci --omit=dev

npm error code 1
npm error path /app/node_modules/better-sqlite3
npm error command failed
npm error command sh -c node-gyp rebuild

npm error gyp info using node-gyp@11.5.0
npm error gyp info using node@22.23.2 | linux | x64
npm error gyp info find Python using Python version 3.11.2 found at "/usr/bin/python3"

npm error gyp ERR! build error
npm error gyp ERR! stack Error: not found: make
npm error gyp ERR! System Linux 6.8.0-136-generic
npm error gyp ERR! node -v v22.23.2
npm error gyp ERR! node-gyp -v v11.5.0
npm error gyp ERR! not ok

ERROR: failed to build: failed to solve:
process "/bin/sh -c npm ci --omit=dev" did not complete successfully:
exit code: 1
```

## Environment

* Branch: latest `master`
* Host kernel: Linux `6.8.0-136-generic`
* Node.js inside the container: `v22.23.2`
* npm: `10.9.8`
* node-gyp: `11.5.0`
* Python: `3.11.2`
* Architecture: `x86_64`

## Working version

The following release builds and runs successfully:

```bash
git clone --branch v1.13.0 --single-branch --depth 1 \
  https://github.com/jo-inc/camofox-browser.git
```

## Possible cause

It appears that a dependency change after `v1.13.0` causes `better-sqlite3` to be compiled from source.

The current Docker build environment includes Python, but does not include the required native build tools, particularly `make` and a C++ compiler.

Installing build dependencies before running `npm ci` may resolve the issue, for example:

```dockerfile
RUN apt-get update \
    && apt-get install -y --no-install-recommends build-essential python3 \
    && npm ci --omit=dev \
    && rm -rf /var/lib/apt/lists/*
```

Alternatively, using a `better-sqlite3` version that provides a compatible prebuilt binary for the selected Node.js version may avoid the source build.

### LINKED ISSUE #8652: Persistence plugin fails to load: Cannot find module 'mcp/lib/cookies.mjs'
## Description

On the Docker image, the `persistence` plugin fails to load at server startup with a module-not-found error, even when it's enabled in `camofox.config.json`.

## Steps to reproduce

```bash
git clone https://github.com/jo-inc/camofox-browser.git
cd camofox-browser
make up
docker logs camofox-browser
```

## Expected behavior

The `persistence` plugin loads successfully and cookies/localStorage are saved to `CAMOFOX_PROFILE_DIR` across restarts.

## Actual behavior

```
{"level":"error","msg":"plugin load failed","plugin":"persistence","error":"Cannot find module '/app/mcp/lib/cookies.mjs' imported from /app/lib/cookies.js"}
```

## Possible cause

`lib/cookies.js` imports from `../mcp/lib/cookies.mjs`:

```js
} from '../mcp/lib/cookies.mjs';
```

but the Dockerfile's `COPY` list only includes `server.js`, `camofox.config.json`, `lib/`, `plugins/`, and `scripts/` — it never copies the `mcp/` directory into the image, so the import fails at runtime in any container build.

Adding `COPY mcp/ ./mcp/` alongside the existing `COPY lib/ ./lib/` line resolves it (only built-in Node modules are needed by `mcp/lib/cookies.mjs`, so no extra `npm ci` step is required for it).

## Environment

* Branch: latest `master`
* Node.js inside the container: `v22.23.2`
* Architecture: `x86_64`

lgwacker:
Two Docker build/runtime issues hit when building `master` fresh:

1. `npm ci --omit=dev` fails compiling `better-sqlite3` via node-gyp because the `node:22-slim` base image has no `make`/C++ toolchain. Adds `build-essential` and `python3` to the existing apt-get install layer (and drops the now-redundant `python3-minimal`, since `python3` already depends on it).
2. The `persistence` plugin fails to load at runtime (`ERR_MODULE_NOT_FOUND`) because `lib/cookies.js` imports `../mcp/lib/cookies.mjs`, but the Dockerfile never copies `mcp/` into the image. Adds `COPY mcp/ ./mcp/` next to the existing `COPY lib/ ./lib/` line. Only built-in Node modules are needed by that file, so no extra install step is required.

Fixes #8639, Fixes #8652

Tests: `make build && make up`, confirmed clean build and `"plugin loaded","plugin":"persistence"` in `docker logs`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,9 @@ ARG ARCH=x86_64
 
 # Install dependencies for Camoufox (Firefox-based)
 RUN apt-get update && apt-get install -y \
+    # Build toolchain for native npm addons (e.g. better-sqlite3)
+    build-essential \
+    python3 \
     # Firefox dependencies
     libgtk-3-0 \
     libdbus-glib-1-2 \
@@ -38,8 +41,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     unzip \
-    # yt-dlp runtime dependency
-    python3-minimal \
     && rm -rf /var/lib/apt/lists/*
 
 # Pre-bake Camoufox browser binary into image (downloaded at build time)
@@ -65,6 +66,7 @@ RUN npm ci --omit=dev
 COPY server.js ./
 COPY camofox.config.json ./
 COPY lib/ ./lib/
+COPY mcp/ ./mcp/
 COPY plugins/ ./plugins/
 COPY scripts/ ./scripts/
 


garfv:
Independent confirmation — both fixes are needed and both work.

Hit exactly these two on a fresh `master` (`af3a2505`, v1.13.1) build on x86_64, Node `v22.23.2`:

1. `npm ci --omit=dev` → `gyp ERR! stack Error: not found: make` on `better-sqlite3`
2. After working around #1, `persistence` loaded as `enabled` in config but died at import with `Cannot find module '/app/mcp/lib/cookies.mjs'`

Worth emphasising how quiet the second one is: the startup log prints `"persistence plugin enabled","profileDir":"/root/.camofox/profiles"` and *then* `plugin load failed`. If you only grep for `enabled` it looks healthy, but **no profile is ever written** — a bind-mounted `~/.camofox` stays empty and every authenticated session is silently lost on restart. That cost me a real login round-trip before I spotted it in the logs.

With `COPY mcp/ ./mcp/` applied, `profiles/<hash>/storage-state.json` is written as expected and an authenticated session survives `docker restart` and a full `docker rm` + `docker run` (external volume).

One optional note on the toolchain layer: `build-essential` is only needed by `npm ci`, so it can be dropped from the runtime image by folding it into that layer:

```dockerfile
RUN apt-get update \
    && apt-get install -y --no-install-recommends build-essential \
    && npm ci --omit=dev \
    && apt-get purge -y build-essential \
    && apt-get autoremove -y \
    && rm -rf /var/lib/apt/lists/*
```

This is what I'm running locally. It keeps the compiler out of a browser image that holds session credentials. Your version is simpler and correct as-is though — purely a size/attack-surface preference, not a blocker.

Note `python3` is still required at runtime for `yt-dlp`, so that part of your change (replacing `python3-minimal`) should stay in the main apt layer regardless.

PR #9252, opened 2026-08-12
fix(vnc): clean up Xvfb socket + lock file on kill()

e-Garcia:
## Problem

`camoufox-js`'s `VirtualDisplay.kill()` (`node_modules/camoufox-js/dist/virtdisplay.js`) sends Xvfb `SIGTERM` but never unlinks the `/tmp/.X11-unix/X<n>` socket or `.X<n>-lock` file it created. Neither `VncVirtualDisplay` (`plugins/vnc/index.js`) nor `DefaultVirtualDisplay` (`server.js`) compensated for that.

On a long-running deployment with the VNC plugin enabled and the browser restarting periodically (health-check-driven restarts, proxy session rotation, etc.), these files accumulate with no cap. On mine, `/tmp/.X11-unix` grew from ~3,200 to ~6,200 stale entries over 6 further days of continuous uptime. That's one new entry per restart cycle, never cleaned up, going back to whenever the VNC plugin was first enabled on that host.

This is a plain resource leak in `/tmp` (harmless on its own until inode/entry limits become a concern on a long-lived host), but it's also user-visible: other local tooling that enumerates `/tmp/.X11-unix` to discover running X displays can pick these up as false candidates. I ran into this indirectly through a remote-desktop tool on the same host intermittently misbehaving, traced it down through the accumulating socket count, and ended up here.

## Root cause

- `VirtualDisplay.execute_xvfb()` spawns Xvfb with an explicit display number and `detached: true`.
- `VirtualDisplay.kill()` only calls `this.proc.kill()` (SIGTERM). It has no cleanup step for the socket/lock file it's implicitly responsible for.
- `_closeBrowserFullyImpl` in `server.js` also runs `_forceKillProcessTree`/`_forceKillBrowserProcesses` shortly after, which can SIGKILL Xvfb as a "browser survivor process" if it hasn't exited from the SIGTERM yet.
- Whether it's the graceful SIGTERM path or the SIGKILL-as-survivor path, nothing ever removes the socket/lock file afterward.

**Caveat on reproduction:** I wrote a minimal isolated repro (spawn/kill `VirtualDisplay` directly in a loop, with and without a connected X client via `xclock`, with and without a follow-up `SIGKILL`) and it did **not** reliably reproduce the leak within a few dozen cycles. Xvfb seems to clean up after itself fine under those simplified conditions. So the trigger seems to depend on something a short synthetic test doesn't capture: sustained real load, a fully-initialized Firefox/Camoufox client rather than a lightweight one, timing under contention, or something else I haven't isolated. I'm flagging this explicitly rather than claiming a clean repro I don't actually have. The leak itself is real and directly confirmed via extended live operation (`ls /tmp/.X11-unix | wc -l` measured over multiple days, not inferred), but I can't hand you a 10-second script that proves the exact trigger condition.

Given that, I think the fix is worth taking regardless of pinning down the precise trigger: nothing in `VirtualDisplay.kill()`'s contract guarantees the socket/lock file is removed, so explicit cleanup after confirmed process exit is a safe, unconditional improvement instead of relying on implicit Xvfb self-cleanup behavior.

## Fix

- Added `removeXvfbDisplayFiles(displayNum, { socketDir, lockDir })` to `lib/tmp-cleanup.js`, alongside the existing `cleanupOrphanedTempFiles`/`cleanupStaleFirefoxProfiles` (same category of problem: files left behind by the Xvfb/Firefox/Camoufox process lifecycle that nothing else cleans up).
- Overrode `kill()` in both `VncVirtualDisplay` and `DefaultVirtualDisplay` to call it once the Xvfb child process has actually exited (`proc.once('exit', ...)` if still running, immediately if it already has). Covers both the plain-SIGTERM and force-killed-as-survivor paths.

## Testing

```
$ NODE_OPTIONS='--experimental-vm-modules' npx jest tests/unit --runInBand --forceExit
Test Suites: 49 passed, 49 total
Tests:       718 passed, 718 total

$ NODE_OPTIONS='--experimental-vm-modules' npx jest --forceExit plugins/
Test Suites: 6 passed, 6 total
Tests:       45 passed, 45 total
```

Added:
- `tests/unit/tmpCleanup.test.js`: 5 new tests for `removeXvfbDisplayFiles` (removes both files, removes whichever exists without touching unrelated displays, no-op when neither exists, no-op for `null`/`undefined` display).
- `plugins/vnc/vnc.test.js`: 3 new tests for `VncVirtualDisplay.kill()`. Covers deferring cleanup until the tracked process's `exit` event, cleaning up immediately if the process already exited, and cleaning up immediately if there's no tracked process at all.

I also ran the full suite (`npm test`, no `--config`) before and after this change to make sure nothing regressed: `tests/e2e/*` fails identically before and after (67 failures either way). Those need `npm run test:e2e`'s global setup (writes `/tmp/camofox-e2e-env.json`), which the plain default config doesn't provide, and that's unrelated to this change. This PR adds exactly 8 new passing tests on top of that baseline (764 → 772 passing), nothing else moved.

## Notes

- Followed `CONTRIBUTING.md`: no `process.env`/`child_process` added to either plugin/route-handler file, no comments narrating what the code does (kept the one rationale comment on `removeXvfbDisplayFiles` itself, matching the existing doc-comment on `cleanupStaleFirefoxProfiles` right above it).
- Scope kept intentionally narrow. Didn't touch the unrelated pre-existing local changes to `camofox.config.json`/`package-lock.json` in my working tree, they're not part of this fix.

### DIFF
--- a/lib/tmp-cleanup.js
+++ b/lib/tmp-cleanup.js
@@ -88,6 +88,28 @@ export function cleanupStaleFirefoxProfiles({ tmpDir, minAgeMs = 2 * 60 * 1000, 
   return result;
 }
 
+/**
+ * Remove a single Xvfb display's socket + lock file. camoufox-js's
+ * VirtualDisplay.kill() (dist/virtdisplay.js) sends SIGTERM but never
+ * unlinks these, and Xvfb itself doesn't reliably self-clean on exit --
+ * especially when it ends up force-killed a moment later as a process-tree
+ * survivor rather than exiting from the SIGTERM cleanly. Call once the
+ * Xvfb process has actually exited, not right after signaling it.
+ */
+export function removeXvfbDisplayFiles(displayNum, { socketDir = '/tmp/.X11-unix', lockDir = process.env.TMPDIR || os.tmpdir() } = {}) {
+  if (displayNum === null || displayNum === undefined) return;
+  for (const target of [
+    path.join(socketDir, `X${displayNum}`),
+    path.join(lockDir, `.X${displayNum}-lock`),
+  ]) {
+    try {
+      fs.rmSync(target, { force: true });
+    } catch {
+      // already gone or permission denied -- not ours to surface
+    }
+  }
+}
+
 /** Recursively calculate directory size (best effort, fast). */
 function _dirSizeSync(dirPath) {
   let total = 0;
--- a/plugins/vnc/index.js
+++ b/plugins/vnc/index.js
@@ -46,6 +46,7 @@
 
 import { resolveVncConfig, startWatcher } from './vnc-launcher.js';
 import { requireAuth } from '../../lib/auth.js';
+import { removeXvfbDisplayFiles } from '../../lib/tmp-cleanup.js';
 
 export async function register(app, ctx, pluginConfig = {}) {
   const { events, config, log, sessions, VirtualDisplay, safeError } = ctx;
@@ -71,6 +72,17 @@ export async function register(app, ctx, pluginConfig = {}) {
         return patched;
       }
       return args;
+    }
+
+    kill() {
+      const displayNum = this._display;
+      const proc = this.proc;
+      super.kill();
+      if (proc && proc.exitCode === null && proc.signalCode === null) {
+        proc.once('exit', () => removeXvfbDisplayFiles(displayNum));
+      } else {
+        removeXvfbDisplayFiles(displayNum);
+      }
     }
   }
 
--- a/plugins/vnc/vnc.test.js
+++ b/plugins/vnc/vnc.test.js
@@ -34,11 +34,18 @@ jest.unstable_mockModule('../../lib/auth.js', () => ({
   requireAuth: () => (_req, _res, next) => next(),
 }));
 
+const mockRemoveXvfbDisplayFiles = jest.fn();
+jest.unstable_mockModule('../../lib/tmp-cleanup.js', () => ({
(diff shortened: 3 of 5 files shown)


e-Garcia:
Update: found and fixed the actual upstream root cause. `camoufox-js`'s `VirtualDisplay.kill()` never cleaned up its own socket/lock file, the same gap this PR compensates for at the call site here. Turns out the original Python `camoufox` this package ports already fixed exactly this in a merged PR ([daijro/camoufox#652](https://github.com/daijro/camoufox/pull/652)), but `camoufox-js` never picked it up. Filed the port upstream: [apify/camoufox-js#322](https://github.com/apify/camoufox-js/issues/322) (issue) / [apify/camoufox-js#323](https://github.com/apify/camoufox-js/pull/323) (PR).

Doesn't change anything about this PR, it's a real, independently-useful fix regardless of what happens upstream (defense in depth, and there could be a gap before any upstream release lands even if merged). Just flagging so it's clear this isn't guesswork about the root cause anymore, and there's a cleaner long-term fix in flight.

skyfallsin:
Thank you for the detailed Xvfb cleanup investigation. v1.13.2 includes tracked Xvfb socket and lock-file cleanup after the display process exits; your investigation is credited in #9424.

skyfallsin:
Included in v1.14.0 via the release branch implementation. Closing this PR now that the fix has shipped; thank you for the contribution.

PR #9329, opened 2026-08-13
Dockerfile: fix two independent aarch64 build/runtime failures

ibrahiem9:
The image can't be built or run on `linux/arm64` today. Two independent defects stack, so fixing either one alone still fails — I spent a couple of builds thinking I'd found "the" root cause before realising there were two.

## 1. Build time — fails on any Debian release

`npm ci --omit=dev` finds no `better-sqlite3` prebuild matching this node/arch and falls back to `node-gyp rebuild`, which dies on `node:*-slim`:

```
Error: not found: make
```

Fix: install `build-essential` + `python3`, run the install, and purge them in the same layer so the toolchain doesn't land in the final image.

## 2. Runtime — bookworm only, and it doesn't fail at startup

`better-sqlite3` ships a **bundled** `prebuilds/linux-arm64.node` linked against `GLIBC_2.38`. Bookworm ships **2.36**. The module resolves fine, the server starts and reports healthy, and then the **first tab open** throws:

```
ERR_DLOPEN_FAILED: /app/node_modules/better-sqlite3/prebuilds/linux-arm64.node:
  /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found
```

`node:22-trixie-slim` ships glibc **2.41**, which satisfies it.

Worth noting for anyone who tries the obvious workaround first: `npm_config_build_from_source=true` + `npm rebuild --build-from-source` does **not** fix this. The loader prefers the bundled `prebuilds/<platform>-<arch>.node` **file** over anything in `build/Release/`, and the forced build left `.deps` + `obj.target` behind with no linked `.node` — a half-build that looks like a build. Matching the glibc is the reliable fix; the alternative is `rm -rf node_modules/better-sqlite3/prebuilds`, which felt more invasive than a base bump.

## Package rename that comes with trixie

`libegl1-mesa` is dropped in trixie — it's `libegl1` now. `libgl1-mesa-dri`, `libgbm1` and `libasound2` are unchanged (no t64 churn on these).

## What I tested

`linux/arm64` only, on Amazon Linux 2023 / aarch64:

- Image builds clean from a cold cache.
- `/health` returns `{"ok":true,"engine":"camoufox","browserConnected":true,"browserRunning":true,"activeTabs":2,...}` with tabs actually open — i.e. past the point where defect 2 used to throw.
- `ldd --version` in the image reports `2.41`, and `libEGL.so.1` is present.

**I have not built or run this on amd64.** amd64 is unaffected by defect 2 (its prebuild targets an older glibc), so for amd64 this change is a base-image bump plus the toolchain fix from defect 1. If you'd rather not move the base image for everyone, the alternative is a conditional on `TARGETARCH` — happy to rework it that way, or to split the two fixes into separate commits if that's easier to review.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,8 +49,12 @@ RUN apt-get update && apt-get install -y \
 
 # Pre-bake Camoufox browser binary into image (downloaded at build time)
 # Note: unzip returns exit code 1 for warnings (Unicode filenames), so we use || true and verify
+# -f so a 404 fails here instead of writing "Not Found" into the .zip: without it the
+# build dies three commands later on "unzip: cannot find zipfile directory", which
+# points at the archive rather than at the URL that was actually wrong. Note the Linux
+# arm asset is named lin.arm64.zip -- pass --build-arg ARCH=arm64, not aarch64.
 RUN mkdir -p /root/.cache/camoufox \
-    && curl -L -o /tmp/camoufox.zip "https://github.com/daijro/camoufox/releases/download/v${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}/camoufox-${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}-lin.${ARCH}.zip" \
+    && curl -fL -o /tmp/camoufox.zip "https://github.com/daijro/camoufox/releases/download/v${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}/camoufox-${CAMOUFOX_VERSION}-${CAMOUFOX_RELEASE}-lin.${ARCH}.zip" \
     && (unzip -q /tmp/camoufox.zip -d /root/.cache/camoufox || true) \
     && rm /tmp/camoufox.zip \
     && chmod -R 755 /root/.cache/camoufox \
@@ -79,6 +83,12 @@ RUN apt-get update \
 COPY server.js ./
 COPY camofox.config.json ./
 COPY lib/ ./lib/
+# lib/cookies.js is a compatibility re-export from ../mcp/lib/cookies.mjs, so mcp/
+# must ship even though the MCP server itself is not run here. Without it the
+# persistence plugin dies at load with ERR_MODULE_NOT_FOUND, the server starts
+# anyway, /health keeps reporting ok, and no profile is ever written -- i.e. the
+# container silently loses the durable-profile feature it exists to provide.
+COPY mcp/ ./mcp/
 COPY plugins/ ./plugins/
 COPY scripts/ ./scripts/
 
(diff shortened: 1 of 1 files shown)


ibrahiem9:
Pushed a third fix (`d1e31fe`) — this one is more serious than the two arm64 build failures, and it affects **every architecture**, not just arm.

## `mcp/` is never copied into the image, so the persistence plugin never loads

`lib/cookies.js` is a compatibility re-export from `../mcp/lib/cookies.mjs`. The Dockerfile copies `server.js`, `lib/`, `plugins/` and `scripts/` — but not `mcp/`. So `/app/mcp` doesn't exist in the built image:

```
"level":"error","msg":"plugin load failed","plugin":"persistence",
"error":"Cannot find module '/app/mcp/lib/cookies.mjs' imported from /app/lib/cookies.js"
```

The server starts anyway and `/health` returns `ok:true` from then on, so the container looks completely healthy while running with **no durable profile** — every session starts cookie-less and logins don't survive a restart. That one `error` line at boot is the only signal. We ran it for four days before noticing the mounted profile directory was 0 bytes.

Fix is `COPY mcp/ ./mcp/` — 7 files, 100K, and shipping it doesn't run the MCP server.

## Verified end to end

Not just "the file is in the image" — I drove a real session and confirmed state reached disk:

```
plugin loaded  plugin=persistence
persistence plugin enabled  profileDir=/root/.camofox/profiles
tab created  url=https://example.com/          [set cookie persistprobe=alive]
tab reaped (inactive)  idleMs=300000
session empty after tab reaper, closing
storage state persisted  reason=tab_reaper_empty_session
```

```json
{"cookies":[{"name":"persistprobe","value":"alive","domain":"example.com","path":"/",
             "expires":1787026826,"httpOnly":false,"secure":false,"sameSite":"None"}],
 "origins":[]}
```

## Also in this commit: `curl -f` on the Camoufox download

Without `-f`, a 404 body gets written into `camoufox.zip` and the build dies three commands later on `unzip: cannot find zipfile directory` — pointing at the archive instead of the URL that was actually wrong. The trailing `test -f camoufox-bin` does catch it, so no broken image ships; the error is just misleading.

I hit this myself by passing `ARCH=aarch64`. The Linux arm asset is published as **`lin.arm64.zip`**, so the correct build arg is `arm64`. Added a comment since it's easy to guess wrong.

## Two things I noticed but did **not** change

- **Checkpoints only happen on session close** (`session:destroying`, `server:shutdown`), so a profile appears ~5 minutes after last activity via the idle reaper, not continuously. Reasonable design; just non-obvious.
- **`docker stop` doesn't checkpoint.** The container exits **137 (SIGKILL)** — node runs as pid 8 under `sh -c` and never receives SIGTERM, so the `server:shutdown` hook never fires. The reaper covers normal operation, but a restart within 5 minutes of activity loses that session's state. That's a design call for you rather than something I wanted to change unasked in this PR.

Still tested on `linux/arm64` only. The `mcp/` fix is architecture-independent, so amd64 users are getting the broken persistence too.

skyfallsin:
Thank you for the ARM64 Docker and MCP reliability work. It has been incorporated into the v1.13.2 release branch with attribution and is included in #9424.

PR #9354, opened 2026-08-14
fix(server): pin browser launch env to X11 so Wayland desktops never see a window

LEAN-EnLAN:
## Problem

On Wayland desktops (Hyprland, GNOME, KDE), the browser opens a **visible window on the user's real compositor** instead of rendering into its private Xvfb virtual display. On tiling compositors it lands in the active workspace, splits the layout, and kicks neighboring windows out of fullscreen.

## Root cause

Firefox selects its display backend **before** honoring `DISPLAY`. The server child inherits `WAYLAND_DISPLAY` from the session (and desktops like omarchy export `MOZ_ENABLE_WAYLAND=1` session-wide, including into the systemd user manager). camoufox-js only overrides `DISPLAY` on the inherited environment, so the Wayland backend wins, `DISPLAY=:N` (the Xvfb) is ignored, and the Xvfb renders for nobody.

Verified live on Hyprland: the camoufox process had `DISPLAY=:153` (correct Xvfb) **and** `WAYLAND_DISPLAY=wayland-1` + `MOZ_ENABLE_WAYLAND=1`; `hyprctl clients` showed the browser as a native Wayland window (`xwayland: false`) tiled into the user's workspace.

## Fix

`launchBrowserInstance` now passes a sanitized environment to `launchOptions` whenever a virtual display is in use (new `lib/display-env.js`): removes `WAYLAND_DISPLAY` and `XDG_SESSION_TYPE`, forces `MOZ_ENABLE_WAYLAND=0` and `GDK_BACKEND=x11`. Playwright **replaces** the child env with the provided one (`options.env ? envArrayToObject(options.env) : process.env`), so the removal is genuine. Side benefit: camoufox-js no longer mutates the server's own `process.env.DISPLAY` (it writes `DISPLAY` onto whatever env object it receives).

Verified on the same machine after the fix: browser child env shows `DISPLAY=:1xx`, no `WAYLAND_DISPLAY`, `MOZ_ENABLE_WAYLAND=0`, `GDK_BACKEND=x11`; `hyprctl clients` lists no camoufox window; screenshots through the API keep working (WebGL via Xvfb intact).

## Also in this PR

- **Test script alignment**: plain `npm test` always failed locally because the default jest config matched `tests/e2e/` without the e2e globalSetup (`ENOENT /tmp/camofox-e2e-env.json`). The default config now ignores `tests/e2e/` and `test:e2e` passes `--config jest.config.e2e.cjs` — the exact split CI already uses (`.github/workflows/ci.yml` never runs plain `npm test`). No CI behavior change.
- **Docs**: new README section (*Virtual display and Wayland desktops*) documenting this failure mode and the related Xvfb `-displayfd` display-number collision hazard on Wayland hosts.

## Tests

- New `tests/unit/displayEnv.test.js` (5 tests: Wayland vars removed, X11 forced, unrelated vars preserved, input never mutated, `process.env` default never mutated).
- Full suite: 46 suites passed, 2 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -662,6 +662,18 @@ Sessions auto-expire after 30 minutes of inactivity. The browser itself shuts do
 
 When a session's tab limit is reached, the oldest/least-used tab is automatically recycled instead of returning an error -- so long-running agent sessions don't hit dead ends.
 
+### Virtual display and Wayland desktops
+
+On Linux the browser is launched headful inside a private Xvfb display rather than in true headless mode: Xvfb provides GLX, so WebGL works via Mesa llvmpipe instead of returning "no context" (a strong bot signal). The window is real but exists only on the virtual display -- nothing should ever appear on your desktop.
+
+Two host-environment leaks can break that invisibility, and both are handled:
+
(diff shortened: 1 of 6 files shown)


LEAN-EnLAN:
Closing — this will land through a staging fork first.

PR #9425, opened 2026-08-19
fix: avoid fixed geolocation in direct sessions

Xipong:
## Summary

- stop assigning fixed San Francisco coordinates to direct (no-proxy) browser contexts
- make direct-session locale, language list, and timezone configurable through explicit `CAMOFOX_*` environment variables
- preserve existing proxy GeoIP behavior, including geolocation permission for proxy-backed contexts
- apply the same direct identity settings to the Google health probe
- centralize direct/proxy identity options in a behavior-tested helper

## Motivation

Every direct connection currently receives San Francisco coordinates and a granted geolocation permission, even when the network exit is elsewhere. That creates a location mismatch and grants a sensitive permission without a caller request.

The change keeps the current `en-US` / `America/Los_Angeles` defaults for compatibility, but no longer fabricates exact coordinates for direct sessions. Operators can align the browser identity with their real exit using:

- `CAMOFOX_LOCALE`
- `CAMOFOX_LOCALES`
- `CAMOFOX_TIMEZONE`

Proxy-backed sessions continue to rely on Camoufox GeoIP and retain their previous geolocation permission behavior.

## Verification

- `npm test -- --runInBand tests/unit/config.test.js tests/unit/browserIdentity.test.js tests/unit/launchCompat.test.js`
- `npm run build`
- `npm run test:mcp`
- CI-equivalent unit + plugin suite: 59 suites / 792 tests passed
- `node --check server.js`

The regression tests were added before the implementation and failed against the previous hard-coded direct-session behavior.

## Scope

This intentionally does not change proxy selection, virtual-display ownership, persistence, health-probe viewport sizing, or shutdown behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ When a proxy is configured:
 - All traffic routes through the proxy
 - Camoufox's GeoIP automatically sets `locale`, `timezone`, and `geolocation` to match the proxy's exit IP
 - Browser fingerprint (language, timezone, coordinates) is consistent with the proxy location
-- Without a proxy, defaults to `en-US`, `America/Los_Angeles`, San Francisco coordinates
+- Without a proxy, direct contexts use the configured locale and timezone and do not set geolocation permissions or coordinates
 
 ### Telemetry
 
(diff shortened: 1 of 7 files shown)


skyfallsin:
Closing in favor of #10667. This branch removes the fixed geolocation, but its default direct-session locale/timezone remain US-Pacific, so it does not resolve the reported egress-IP mismatch. A replacement should define the intended direct-session identity policy and be rebased on current main.

PR #9436, opened 2026-08-20
fix: stop broadcasting typed text to plugins

shauneccles:
### Problem

`server.js` emits the `tab:type` plugin event with the typed text verbatim:

```js
pluginEvents.emit('tab:type', { userId, tabId, text: req.body.text, ref, mode });
```

and `lib/plugins.js` documents `text` as part of the public plugin contract. The text typed into a login form is a password, so any plugin subscribing to `tab:type` receives credentials it never asked for — including plugins installed from elsewhere.

No shipped plugin subscribes to `tab:type`, so nothing in-tree depends on the field today.

### Change

Emits `textLength` instead of `text`. The payload is built by a small pure helper placed next to the contract it implements in `lib/plugins.js`, so the shape is unit-testable without spawning a server — matching the approach in `tests/unit/accessKey.test.js`.

### Contract change

This breaks the documented plugin event: a plugin reading `event.text` now reads `undefined`. `lib/plugins.js` is updated to document `textLength`. Happy to gate it behind config instead if you'd rather not change the contract in a patch release.

### Tests

`tests/unit/pluginTypeEvent.test.js` — 4 tests, one asserting no property of the serialised payload contains the typed text.

```
tests/unit/{pluginTypeEvent,plugins,sessionDestroyingEvent,serverShutdownEvent}.test.js
Test Suites: 4 passed    Tests: 31 passed
```

For transparency: the full `tests/unit` suite is already red on `master` at `e5a36f5` — 10 suites / 17 tests fail before this change. None of those suites import `lib/plugins.js`.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -36,7 +36,7 @@
  *
  *   INPUT
  *     tab:click               { userId, tabId, ref, selector }
- *     tab:type                { userId, tabId, text, ref, mode }
+ *     tab:type                { userId, tabId, textLength, ref, mode }
  *     tab:scroll              { userId, tabId, direction, amount }
  *     tab:press               { userId, tabId, key }
  *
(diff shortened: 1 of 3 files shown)


skyfallsin:

PR #9522, opened 2026-08-23
fix(ci): keep upstream release metadata out of shell

shauneccles:
### Problem

The mirror workflow reads `tag_name` and `html_url` from `daijro/camoufox` release JSON, writes them to step outputs, then interpolates those outputs directly into three Bash `run:` blocks. GitHub substitutes expressions before Bash parses the script, including inside the unquoted release-notes heredoc.

That gives third-party release metadata a shell parsing boundary in a job with `contents: write`.

### Change

- pass all release outputs through step-level environment variables
- reference only quoted shell variables inside `run:` blocks
- move `contents: write` from workflow scope to the mirror job

The release names, notes, manifest and asset checks are unchanged.

### Verification

- workflow YAML parsed successfully
- default zizmor persona: no findings
- pedantic zizmor: only the pre-existing low-severity concurrency recommendation remains
- `git diff --check` passed


skyfallsin:

PR #9530, opened 2026-08-23
build: pin and verify yt-dlp

shauneccles:
### Problem

The production image downloads `yt-dlp` from `releases/latest` without checking its content. A rebuild of the same source can therefore install a different executable, and a failed download is not rejected until a later command.

### Change

- pin yt-dlp to release `2026.08.19`
- pin the release asset SHA-256 published by GitHub
- use `curl -fL` so HTTP failures stop the build
- verify the downloaded file before making it executable

Both values are build arguments so a future update remains one reviewed version-and-digest change.

### Verification

- GitHub release API reports digest `sha256:1fa6733c37ea6fb51c99ad8fe785e7b7e5f3246c9b980230329d4fb72ed8d4d6` for the `2026.08.19` `yt-dlp` asset
- downloaded the exact asset and verified the checksum locally
- `git diff --check` passed


skyfallsin:

PR #9613, opened 2026-08-27
fix: persist completed download artifacts

steveonjava:
## What this does

Persist completed browser downloads as short-lived artifacts that callers can retrieve after the download event. Failed downloads do not leave an artifact that can be read later.

## Why

Download records previously pointed at temporary files and did not provide a stable retrieval path. Blob downloads also need browser-side decoding before the server can persist their bytes.

## Changes

- Store completed downloads under a private temporary root with a 15-minute expiry.
- Add size limits, SHA-256 metadata, MIME detection, safe filenames, ownership checks, and artifact content reads.
- Remove artifacts when Playwright reports a download failure or when records expire.
- Add OpenAPI entries and unit coverage for failure cleanup, blob metadata, content ownership, expiry, size limits, and safe filenames.

## Verification

- Focused download tests: 23 passed.
- Full unit suite: 780 passed; 29 failures require the unavailable Camoufox cache and binaries.
- `npm run build`: passed.
- `npm run generate-openapi`: produced no diff.
- `git diff --check`: passed.

## Files changed

- `lib/downloads.js` - artifact persistence, cleanup, metadata, and content reads.
- `server.js` - download artifact routes and response wiring.
- `openapi.json` - generated artifact endpoint documentation.
- `tests/unit/downloads.test.js` - download artifact and failure-cleanup coverage.

## Related

Refreshed the prior-art snapshot at packaging time. No overlapping open PRs or issues were found.


steveonjava:
Rebased this onto current `master` and resolved the `lib/downloads.js` conflict without dropping the newer fetch-current-resource PDF path. Fetched resources now use the same owned, expiring artifact store as browser downloads, with a regression test for owner isolation.

Local verification: `tests/unit/downloads.test.js` (24 passing), `npm run build`, `npx tsc -p . --noEmit`, and regenerated `openapi.json`. GitHub created fresh CI and ClawHub runs for `c058f7f`, but both are waiting for repository approval before any jobs can start.

PR #9778, opened 2026-08-29
docs(persistence): document the on-demand checkpoint and correct the trigger list

shauneccles:
## What's confusing today

Reading the persistence plugin's docs, there is no way to find out how to force a checkpoint.

`plugins/persistence/README.md` lists the checkpoint triggers as "Cookie import / session close / shutdown", and `plugins/persistence/AGENTS.md` lists `session:cookies:import` / `session:destroyed` / `server:shutdown`. Neither mentions `session:storage:export`, which the plugin does listen for (`plugins/persistence/index.js:148`) and which is the only user-callable way to force a checkpoint.

The reason it is easy to miss is that the route emitting that event, `GET /sessions/:userId/storage_state`, is registered by the **vnc** plugin (`plugins/vnc/index.js:137`, emitting at `:160`), not by the persistence plugin — which registers only the `DELETE` verb on the same path (`plugins/persistence/index.js:178`). `plugins/vnc/README.md` documents the pairing from the vnc side, but a reader who starts at the persistence plugin has no thread to pull.

This matters for the default configuration in particular: the shipped `camofox.config.json` has `"persistence": { "enabled": true }` alongside `"vnc": { "enabled": false }`, so out of the box there is no on-demand checkpoint available at all, and storage state reaches disk only on cookie import, session close and shutdown. That is reasonable behaviour, it just isn't written down anywhere the reader will look.

## Changes

Documentation only — no code, no behaviour change.

**`plugins/persistence/README.md`**
- Add an **On demand** bullet covering `GET /sessions/:userId/storage_state`, the `session:storage:export` event, and the fact that the route belongs to the vnc plugin so it needs vnc enabled.
- Note that the shipped config leaves vnc disabled, so the default configuration has no on-demand checkpoint.
- Mention that first-run bootstrap cookies are also checkpointed (`index.js:133`), which the trigger list previously omitted.

**`plugins/persistence/AGENTS.md`**
- Add `session:storage:export` to the trigger list, and note where it is emitted from.
- Correct `session:destroyed` to `session:destroying`. The checkpoint runs at `index.js:155`, while the context is still alive; `session:destroyed` (`index.js:166`) only clears tracking. The distinction is the whole point of the event that #75 added, so it seemed worth getting right in the guide.
- Correct `storage_state.json` to `storage-state.json`, matching `lib/persistence.js:17`. The *route* path keeps its underscore; only the on-disk filename is hyphenated, which is exactly the sort of thing that is easy to trip over.

## Not in scope

This does not revisit timer-based checkpointing. #7223 settled that question and the reasoning there — that a timer serialises even when storage hasn't changed — makes sense. This PR only makes the mechanism that already exists discoverable from the plugin that depends on it.

## Verification

- `npx jest plugins/persistence` — 2 suites, 16 tests, all passing (unchanged; the diff touches no code).
- Relative link `../vnc/` resolves from `plugins/persistence/`.
- Confirmed no remaining `storage_state.json` in the plugin docs, and that both `storage_state` route references are untouched.

Verified against `e5a36f5`. Thanks for the plugin — the hashed per-user directories and the atomic tmp-write + rename in `lib/persistence.js` have been solid.


shauneccles:
Pushed a second commit (`dc3c2e4`) covering one more inaccuracy in the same file, so anyone who has already reviewed only needs to look at that one.

`plugins/persistence/AGENTS.md` described `index.js` in its key files list as "lifecycle hooks (no routes, no `child_process`)", but the plugin does register `DELETE /sessions/:userId/storage_state` (`index.js:178`) — and the "How It Works" list a few lines above already documents that route, so the file contradicted itself. The entry now names the route.

I kept the `no child_process` half, which is still accurate: `index.js` imports only `node:fs/promises`, `lib/persistence.js` and `lib/cookies.js`, and nothing in the plugin spawns a process.

Still documentation only — the diff is two `.md` files, no code. `npx jest plugins/persistence` passes at 16/16 (unchanged, as expected).

PR #9819, opened 2026-08-29
Fix SESSION_TIMEOUT_MS=0 and BROWSER_IDLE_TIMEOUT_MS=0 (0 = never)

audinue:
## Summary

Fixes the handling of \`SESSION_TIMEOUT_MS\` and \`BROWSER_IDLE_TIMEOUT_MS\` when their value is \`0\`.

### Before
- \`BROWSER_IDLE_TIMEOUT_MS=0\` scheduled \`setTimeout(..., 0)\`, which shut the browser down **immediately** when idle — the opposite of the documented \`0 = never\`.
- \`SESSION_TIMEOUT_MS=0\` collapsed to the \`600000\` default via \`parseInt(...) || 600000\`, so sessions still expired instead of never.

### After
- \`0\` now explicitly means **disabled / never** for both timeouts.
- Unset env vars still fall back to their defaults (session: \`600000\`, idle: \`300000\`).
- \`scheduleBrowserIdleShutdown()\` skips scheduling when \`BROWSER_IDLE_TIMEOUT_MS <= 0\`.
- The session-cleanup interval skips expiry when \`SESSION_TIMEOUT_MS <= 0\`.
- README updated so \`SESSION_TIMEOUT_MS\` documents \`(0 = never)\` to match \`BROWSER_IDLE_TIMEOUT_MS\`.

### Verification
- \`node --check\` passes on \`lib/config.js\` and \`server.js\`.
- Parsing logic tested for unset / \`0\` / default-docs / custom values.

🤖 Generated with [OpenCode](https://opencode.ai)


skyfallsin:

Issue #8559, opened 2026-07-31
page.goto to bare image URLs always times out (waitUntil domcontentloaded never fires for image-only responses) and triggers session destroy


jamesHGBlackSmith:
## Summary
`page.goto(url, { waitUntil: 'domcontentloaded', timeout: 30000 })` **never resolves** when navigating to a bare image URL (image-only response, e.g. `https://p16-common-sign.tiktokcdn.com/...jpeg` or any direct `.jpg/.png` URL). The navigation always times out after 30s, even though the same URL loads in ~1.2s via curl and via `fetch()` from inside the page.

## Repro (verified 2026-07-31, Camofox 1.11.2 + 1.13.0)
1. `POST /tabs` with any URL (e.g. example.com) — succeeds
2. `POST /tabs/:tabId/navigate` with a bare image URL:
   - `https://p16-common-sign.tiktokcdn.com/tos-maliva-i-photomode-us/e0627ada4f9b4557bbf31e776d555847~tplv-photomode-image.jpeg?dr=14555&x-expires=1785682800&x-signature=s1bsg8HqaXOIvr9UoODZOOlv2cw%3D&t=4d5b0474&ps=13740610&shp=81f88b70&shcp=9b759fb9&idc=my&ftpl=1`
   - or even `https://images.unsplash.com/photo-1504674900247-0877df9cc836?w=200` (non-TikTok domain)
   → always `{"error":"navigate timed out after 30000ms"}`

Control tests that PASS:
- `curl` to the same URL: HTTP 200 in ~1.2s (both HTTP/1.1 and HTTP/2, with and without browser-like headers)
- `fetch()` from inside the page (via evaluate): HTTP 200, `content-type: image/jpeg`, full 2MB body
- `page.goto` to normal HTML pages (example.com, google.com, tiktok.com): fine

So the network/CDN is not the problem — `domcontentloaded` simply never fires for image-only document navigation in Camoufox.

## Impact (severe for parallel agents)
When the navigate times out, the server logs `navigation timeout — destroying session for fresh proxy` and destroys the **entire shared user session** (`destroying dead session`). Since all parallel subagents share the same `userId`, ONE subagent navigating to a bare image kills ALL concurrent tabs → every other agent gets 404 `Tab not found` for the rest of its run. This cascades hard in fan-out scenarios.

## Suggested fix
[...]

PR #9821, opened 2026-08-29
fix: only destroy a session on navigation timeout when a proxy can rotate

### LINKED ISSUE #9820: Navigation timeout destroys the whole session even with no proxy to rotate, killing concurrent tabs
## Summary

A navigation timeout destroys the entire user session even when no proxy is configured — so every other tab under that `userId` dies for a rotation that cannot happen.

`handleRouteError` in `server.js` has two adjacent recovery branches. The proxy-error branch is guarded on there being a proxy to rotate to:

```js
if (isProxyError(err) && proxyPool?.canRotateSessions && userId) {
```

The navigation-timeout branch immediately below it is guarded on nothing:

```js
const NAVIGATION_TIMEOUT_ACTIONS = new Set(['click', 'navigate', 'open_url']);
if (isTimeoutError(err) && err.code !== 'tab_timeout' && userId && NAVIGATION_TIMEOUT_ACTIONS.has(action)) {
  ...
  destroySession(userId).catch(() => {});
}
```

…despite its own comment giving proxy poisoning as the entire reason, and #2472 introducing it on exactly that basis:

> **Fix 2: Navigation timeout → session destruction.** The browser context shares a single proxy session, so one poisoned page (e.g. Cloudflare holding connection for 30s) kills all subsequent navigations.

That reasoning is sound where a proxy pool exists. With `proxyMode: null` there is no proxy session to poison and nothing to get fresh, so the teardown has no upside and a large downside.

## Impact

`sessionKey` partitions tabs, but the browser context, cookie jar and profile are keyed on `userId` — so concurrent callers sharing a `userId` share a session. One caller's timeout therefore takes every other caller's tabs, and their next request gets `404 Tab not found`.

This is the blast radius reported in #8559 ("ONE subagent navigating to a bare image kills ALL concurrent tabs"). That issue was closed with v1.14.0, which fixed the *trigger* it described — direct image navigation now completes at commit — but not the radius. Any other navigation timeout still takes the whole session: a slow page, a hung connection, an unroutable host.

The reporter of #8559 suggested two fixes and only the first was taken. This is the narrow form of the second.

## Reproduction

Run with no proxy configured (`proxyMode: null`), then under a single `userId`:

1. Open two tabs, A and B, on any working sites.
2. Navigate tab A to an address that will hang — a blackholed IP such as `http://10.255.255.1/` works.
3. After the 30s navigation timeout, tab B is gone. Any call against it returns `404 Tab not found`.

Observed on 1.14.0. In our deployment this happened for real: an agent navigating to a slow page destroyed a human's concurrent noVNC browsing session.

## Suggested fix

Guard the branch the same way the proxy-error branch above it is guarded — `proxyPool?.canRotateSessions` — and fall through to the existing per-tab consecutive-timeout reaper when there is no proxy. That reaper is the right granularity for a stuck tab when no proxy is involved, and still collects one after `MAX_CONSECUTIVE_TIMEOUTS`. `recordNavFailure` should be kept on both paths so session health tracking is unchanged.

Behaviour with a proxy pool configured is completely unchanged.

PR follows.

shauneccles:
Fixes #9820.

## The problem

`handleRouteError` destroys the entire user session on a navigation timeout, unconditionally. The branch immediately above it — proxy errors — is guarded on there being a proxy to rotate to:

```js
if (isProxyError(err) && proxyPool?.canRotateSessions && userId) {
```

The navigation-timeout branch has no such guard, even though proxy poisoning is the whole reason it exists. #2472 introduced it on exactly that basis:

> **Fix 2: Navigation timeout → session destruction.** The browser context shares a single proxy session, so one poisoned page (e.g. Cloudflare holding connection for 30s) kills all subsequent navigations.

Correct where a proxy pool exists. With `proxyMode: null` there is no proxy session to poison and nothing to get fresh, so the session is destroyed for a rotation that cannot happen.

The cost is not local to the caller. `sessionKey` partitions tabs, but the context, cookie jar and profile are keyed on `userId` — so concurrent callers sharing a `userId` share a session. One caller's timeout takes every other caller's tabs, and their next request gets `404 Tab not found`.

That is the blast radius reported in #8559. v1.14.0 fixed the trigger described there — bare image URLs now complete at commit — but not the radius, so any other navigation timeout still takes the whole session: a slow page, a hung connection, an unroutable host. The reporter suggested two fixes and only the first was taken; this is the narrow form of the second.

## The change

Guard the branch the same way the one above it is guarded, and fall through to the existing per-tab consecutive-timeout reaper when there is no proxy. That reaper is the right granularity for a stuck tab when no proxy is involved, and still collects one after `MAX_CONSECUTIVE_TIMEOUTS`.

`recordNavFailure` is called on both paths, so session health tracking is unchanged either way, and the no-proxy path logs that it declined to tear the session down rather than going silent.

**Behaviour with a proxy pool configured is unchanged.**

## Tests

`tests/unit/navigationTimeout.test.js` gains a `describe` block for the no-proxy case:

- each of `click` / `navigate` / `open_url` spared when sessions cannot rotate
- the navigation failure still recorded (health signal preserved)
- the same timeout **still** destroying the session when a proxy *can* rotate
- proxy errors also requiring a rotatable proxy
- non-navigation timeouts still tracked per-tab, and a stuck tab still collected after 3

The file's `simulateErrorHandling` replica gained a `canRotateSessions` parameter defaulting to `true` — which is what the cases already there assumed — so existing tests are untouched. Its proxy-error branch now carries the same guard the real one has; it had drifted.

## Verification

`npm run test:unit` is **812 passing / 1 failing both before and after** this change.

The one failure is pre-existing on an unmodified checkout of `master`, confirmed by running it on a clean tree: `destroys a tab before releasing queued evaluate work after an evaluate timeout`. It involves no navigation action.

Also verified against a running 1.14.0 browser with no proxy: two tabs under one `userId`, navigate one to a blackholed address, real 30s timeout — before, the second tab was destroyed; after, both survive and the log shows

```
navigation timeout — no proxy to rotate, leaving the session up
```


skyfallsin:

PR #10262, opened 2026-09-04
fix(cookies): map a session expiry onto -1 for addCookies()

nayrosk:
## The bug

Netscape cookie files use `0` in the expiry field to mean "session cookie".
That is the only convention the format has for it, and it is what browser
exports actually contain.

`parseNetscapeCookieFile()` read that field with a bare `Number(parts[4])` and
handed the result to Playwright's `context.addCookies()`. Playwright does not
read `0` as "session": it spells that `-1`, and reads `0` as a Unix timestamp,
so 1970-01-01. The cookie is therefore already expired the moment it is added,
and Playwright drops it.

Nothing surfaces this. The import reports every cookie as imported and the
caller sees a success. The cookie is simply never attached to any subsequent
request, which then looks like a session that was never valid in the first
place. Because session cookies are the common case in an exported file, this
affects ordinary usage rather than an edge case.

## The change

Anything that is not a positive finite number maps to `-1`. That also covers an
unparsable expiry, which previously produced `NaN` and is rejected by
`addCookies()` outright.

## Why the tests changed too

`tests/unit/netscapeParser.test.js` carried a second, hand-written copy of the
parser, with a comment asking that any change to the real one be mirrored here.
That is how a parser and its tests can agree perfectly while both being wrong:
the old suite asserted `expires: 0` for a session cookie and `NaN` for an
unparsable one, so it passed on precisely the broken behaviour.

The copy is gone. The tests import the shipped parser through `lib/cookies.js`,
so the implementation and its tests can no longer drift, and three cases now
cover session, negative and unparsable expiries.

## How it was found

Root-caused end to end while wiring this server into an agent that imports a
logged-in session from a cookie file: the import reported success, and every
later request went out unauthenticated. Verified against the parser directly:
`0` and a garbage value both come back as `-1` now, a real timestamp is left
alone, and `#HttpOnly_` handling is unchanged.


skyfallsin:

PR #10263, opened 2026-09-04
fix(reporter): stop retaining a Request/Response pair per request

nayrosk:
## Symptom

The server was OOM-killed 21 times over four days, a median 3.9 hours apart,
under `--max-old-space-size=512`, always with `FATAL ERROR: Reached heap
limit`. The tab in question was long-lived and served a steady trickle of
requests.

## Cause

`createTabHealthTracker()` attaches `page.on('request')`, `'response'`,
`'requestfinished'` and `'requestfailed'` for the whole life of every tab.

Playwright only constructs its internal `Request`/`Response` dispatcher objects
for a non-navigation request when at least one of those four event names is
subscribed on the page or its browser context
(`lib/server/dispatchers/browserContextDispatcher.js`,
`_shouldDispatchNetworkEvent`). Once created, they are registered as children
of the page's dispatcher and are **never disposed individually**. Only the page
closing releases them. There is an internal `setMaxDispatchersForTest`
threshold of 10,000, but it is not reachable through `playwright-core`'s public
`exports` map and is not scoped per object type, so lowering it would also have
Playwright dispose `JSHandle`/`ElementHandle` objects a script is still using.

So a tab that lives for hours and serves thousands of requests accumulates one
`Request`+`Response` pair, plus their `Map`/`ManualPromise`/closure
bookkeeping, per request, for as long as it stays open.

## Measurement

CDP heap snapshots taken before and after 3000 `fetch()` calls on a single
long-lived tab, with a forced GC between the load and the second snapshot:

| tracker | retained after GC |
|---|---|
| listeners attached (before) | +945k nodes, +60.6 MB shallow |
| listeners not attached (after) | +5.5k nodes, +0.66 MB, returns to baseline |

Bisected against `playwright-core` 1.59.1 as vendored in the image.

## The change

Request-level tracking becomes opt-in through a new `options.trackRequests` on
`createTabHealthTracker(page, options)`, threaded through
`createTabState(page, options)`. It defaults to `false`, so all 8 existing call
sites keep their current behaviour without being touched, and none of them
needs the data.

`crash`, `pageerror` and `dialog` stay attached unconditionally. They are
load-bearing and none of their handlers subscribes to per-request events, so
they do not drive the leak.

`health.redirectStatusCodes` is also bounded to its 20 most recent entries. It
was appended to on every redirect-tagged navigation response and only ever
reset by a non-redirected top-level navigation, which a single-page app that
never performs one again will not trigger, so it grew for the tab's whole life.
That one is independent of the leak above and was found while reading the same
function.

## Tradeoff

With `trackRequests` off, `inflightRequests`, `requestFailures`,
`statusCounts`, `maxRedirectDepth`, `redirectStatusCodes`, `botDetection` and
`lastNavResponseSize` read as zero or empty rather than as unknown, which is a
real false-negative risk for anything that would decide behaviour on
`botDetection`. Nothing in this repository does today. A caller that wants it
can pass `trackRequests: true` for that specific tab and accept the bounded
cost, or read the same information from each navigation's own `page.goto()`
return value, which requires no standing listener.

Happy to split the `redirectStatusCodes` bound into its own commit or drop it
here if you would rather keep this focused on the listeners.


skyfallsin:

PR #10465, opened 2026-09-08
fix: classify aborted navigation before generic network errors

Bortlesboat:
`classifyError(new Error('page.goto: net::ERR_ABORTED at https://example.com/'))` currently returns `network`: the generic `net::` check runs before the existing aborted-navigation check.

Check the specific navigation failures before the network fallback so aborted navigations receive the intended `nav_aborted` diagnostic label. Add coverage showing connection-refused and DNS errors still classify as `network`.

Validation: the new abort assertion fails before the change; all 23 tests in `tests/unit/navigationTimeout.test.js` pass with `NODE_OPTIONS=--experimental-vm-modules` and Jest. This change is limited to error classification; no browser session or live navigation was used for validation.


skyfallsin:

PR #10479, opened 2026-09-08
feat: natural human-like mouse glide before click

Wilsonchoo2017:
## Intent

add a natural human-like mouse glide to the camofox browser /click endpoint so the cursor visibly eases/jitters through a multi-segment path to the target instead of teleporting, reducing Taobao/Shopee anti-bot risk

## What Changed

- Added `lib/humanize.js` with `rand()` and `randomPointInBox()` helpers: pick a randomized aim point inside an element's padded bounding box instead of dead center.
- `/tabs/:tabId/click` now glides the cursor onto the element (best-effort, logged and swallowed on failure so the subsequent Playwright click still reports the real error) before every click attempt, aims the native mouse sequence at the randomized point, and replaces the fixed 50ms waits with randomized 40-120ms pauses. The eased multi-segment path comes from Camoufox's `humanize` expansion of the single `mouse.move`.
- Added unit tests for the aim-point helpers (interior bounds, non-center spread, degenerate zero-size box, `rand` range) and updated `timeoutContinuation.test.js`, which asserted a `withTimeout`-wrapped `mouse.move` per action, to cover only `down`/`up`.

## Risk Assessment

✅ Low: The two prior error-level defects are genuinely fixed (click errors propagate again, Playwright actionability checks are back on the primary path), the glide is a bounded best-effort step that cannot mask a click failure, cursor state is now per-tab, and the new helper has real executable coverage; what remains is one efficacy gap for below-the-fold targets plus minor unused knobs.

## Testing

Ran the new humanize unit test and the click-related unit tests (all pass except a pre-existing, host-speed-related failure in timeoutContinuation that also fails on the base commit), then drove the actual endpoint end-to-end against a live Camoufox browser: a local page recorded every mousemove during POST /tabs/:id/click, showing a 229-241 point eased and jittered path into the target with no jump over 27px and the click landing 33-50px off dead center, varying between runs. The same runs confirm the round-1 timeout regression is fixed (200 in ~11s at the default 30s handler budget, where the previous head returned 410 tab_timeout). I also verified the glide's best-effort error swallowing does not hide real failures (missing selector and display:none both return 422 element_not_actionable) and that an overlay-covered button still resolves via the pre-existing force-click fallback exactly as on base. Visual evidence is a screenshot of the recorded cursor path drawn over the rendered page.

- Evidence: Cursor path recorded by the page during POST /click, drawn over the rendered page (red = path, green ring = where the click landed, black dot = old dead-center target) (local file: <code>~/.no-mistakes/evidence/01M1ZFFCW3WZA6HTPP9Z3DBN4P/glide-path-after.png</code>)
- Evidence: Second run of the same click, showing the aimed point differs between runs (local file: <code>~/.no-mistakes/evidence/01M1ZFFCW3WZA6HTPP9Z3DBN4P/glide-path-run2.png</code>)

<details>
<summary>Evidence: End-to-end /click transcript: glide metrics, off-center landing, and failure propagation</summary>

<code>--- after ---&#10;POST /tabs/:id/click -&gt; HTTP ok={&#34;ok&#34;:true,...} in 11044ms (HANDLER_TIMEOUT_MS default 30000)&#10;page-observed mousemove events during the click: 229&#10;cursor path length: 988px across 228 segments, largest single jump 13.6px&#10;button rect: x=560 y=430 w=220 h=64, dead center = (670,462)&#10;click landed at: (637,449) offset from dead center: dx=-33.0 dy=-13.0&#10;&#10;--- run2 ---&#10;POST /tabs/:id/click -&gt; HTTP ok=... in 11196ms&#10;page-observed mousemove events during the click: 241&#10;cursor path length: 1079px across 240 segments, largest single jump 27.0px&#10;click landed at: (720,453) offset from dead center: dx=50.0 dy=-9.0&#10;&#10;--- failure propagation ---&#10;click on a selector that matches nothing -&gt; HTTP 422 element_not_actionable&#10;click on a display:none element -&gt; HTTP 422 &#34;Element not visible (no bounding box)&#34;</code>

```text
--- after ---
POST /tabs/:id/click  -> HTTP ok={"ok":true,"url":"http://127.0.0.1:60070/","refsAvailable":true} in 11044ms (HANDLER_TIMEOUT_MS default 30000)
page-observed mousemove events during the click: 229
cursor path length: 988px across 228 segments, largest single jump 13.6px
first point: (1,1)   last point: (637,449)
button rect: x=560 y=430 w=220 h=64, dead center = (670,462)
click landed at: (637,449)  offset from dead center: dx=-33.0 dy=-13.0
screenshot: ~/.no-mistakes/evidence/01M1ZFFCW3WZA6HTPP9Z3DBN4P/glide-path-after.png

--- run2 ---
POST /tabs/:id/click  -> HTTP ok={"ok":true,"url":"http://127.0.0.1:60165/","refsAvailable":true} in 11196ms (HANDLER_TIMEOUT_MS default 30000)
page-observed mousemove events during the click: 241
cursor path length: 1079px across 240 segments, largest single jump 27.0px
first point: (16,21)   last point: (720,453)
button rect: x=560 y=430 w=220 h=64, dead center = (670,462)
click landed at: (720,453)  offset from dead center: dx=50.0 dy=-9.0
screenshot: ~/.no-mistakes/evidence/01M1ZFFCW3WZA6HTPP9Z3DBN4P/glide-path-run2.png

--- blocked click ---
click on button covered by a full-screen overlay -> UNEXPECTED 200 {"ok":true,"url":"http://127.0.0.1:60322/","refsAvailable":true}  (11486ms)
same click after removing the overlay      -> HTTP 200 {"ok":true,"url":"http://127.0.0.1:60322/","refsAvailable":true}

--- failure propagation ---
click on a selector that matches nothing  -> HTTP 422 {"error":"Element not actionable: no bounding box within 3000ms (element likely detached after page change). Call snapshot to refresh refs and retry.","retryable":true,"code":"element_not_actionable","recovery":"snapshot_then_retry"}
click on a display:none element         -> HTTP 422 {"error":"Element not visible (no bounding box)","retryable":true,"code":"element_not_actionable","recovery":"snapshot_then_retry"}
```
</details>
<details>
<summary>Evidence: Before/after latency comparison at the product-default handler timeout</summary>

<code>HANDLER_TIMEOUT_MS = 30000 (product default):&#10;base e5a36f5 : 200 ok in 9746ms / 7987ms | mousemoves: 111,113&#10;head e803440 : 410 tab_timeout (round-1 regression)&#10;head 668c2f3 : 200 ok in 11044ms / 11196ms | mousemoves: 229 / 241, longest jump 13.6px / 27.0px</code>

```text
camofox /click - before vs after (same page, same script, real Camoufox browser)

HANDLER_TIMEOUT_MS = 60000 (raised, so the click can finish):
  base e5a36f5 : 5776ms / 7526ms  | mousemoves seen by page: 111,113 | click handler fired: 2x (double click)
  head e803440 : 25282ms / 27076ms / 27124ms | mousemoves: 616,645,611 | click handler fired: 1x

HANDLER_TIMEOUT_MS = 30000 (product default, lib/config.js:111):
  base e5a36f5 : 200 ok in 9746ms / 7987ms
  head e803440 : 410 {"error":"action timed out after 30000ms","code":"tab_timeout","recovery":"create_new_tab"}  (tab destroyed)
  head e803440 : 410 {"error":"native mouse move timed out after 1ms","code":"tab_timeout"}

Cause: camoufox is launched with humanize:true (server.js:1125), so every page.mouse.move is
itself expanded into a slow humanized micro-path. The 8-step glide issues 9 such moves per click.

--- re-measured on 668c2f3 (single humanized glide move, JS stepping dropped) ---
HANDLER_TIMEOUT_MS = 30000 (product default):
  base   e5a36f5 : 200 ok in 9746ms / 7987ms  | mousemoves seen by page: 111,113 (from the fallback sequence only)
  head   e803440 : 410 tab_timeout (regression, now fixed)
  head   668c2f3 : 200 ok in 11044ms / 11196ms | mousemoves: 229 / 241, longest single jump 13.6px / 27.0px
                   click landed 33px / 50px off the element's dead center (varies per run)
Round-1 timeout regression is resolved; the glide is still a continuous multi-segment path.
```
</details>
- Outcome: ⚠️ 1 info across 2 runs (45m9s)

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 4 issues (2 warnings, 2 infos)</summary>

- 🚨 `server.js:3649` - /click now returns 200 {ok:true} when no click actually happened. The three fallback branches that previously ended in `await dispatchMouseSequence(locator)` (which threw on failure) were replaced with a bare `log(&#39;warn&#39;, ...)` and no rethrow: server.js:3634 (Google SERP force-click failure), server.js:3649 (overlay-intercept force-click failure), server.js:3652 (not-visible/timeout). Concrete sequence: element covered by a Taobao login overlay -&gt; dispatchMouseSequence throws (or clicks nothing useful) -&gt; caught and logged at 3627 -&gt; `click({timeout:3000})` throws &#39;intercepts pointer events&#39; -&gt; force click also fails -&gt; only a warn is logged -&gt; doClick resolves -&gt; route replies ok:true. The caller believes the click landed and proceeds. Fix: rethrow forceErr / err in those branches so the route still surfaces 422/500 as before.
- 🚨 `server.js:3624` - dispatchMouseSequence was promoted to the PRIMARY path for every click (server.js:3624), so ordinary clicks now dispatch raw mouse.down/up at coordinates from locator.boundingBox() with no actionability check. boundingBox() reports the target&#39;s geometry even when another element is stacked on top, so when a cookie banner / login modal / sticky header covers the target, the synthetic press lands on the overlay, the sequence completes without error, and the route reports ok:true having clicked the wrong element - exactly the silent wrong-result class Playwright&#39;s &#39;intercepts pointer events&#39; check existed to prevent. The intent only requires that the cursor visibly glides to the target; a narrower form satisfies it: run _naturalMouseMove to the target point first, then keep the existing locator.click() ordering (normal -&gt; force -&gt; mouse sequence) so obscured elements are still detected. Remedy changes deliberate ordering, so it needs the author&#39;s call.
- ⚠️ `server.js:4085` - Per-keystroke delay jitter in the /type keyboard mode (server.js:4085-4091) is outside the stated intent, which covers &#39;the camofox browser /click endpoint&#39;. No intent requirement needs it; recommend removing it from this change and shipping it separately. It also carries its own defect: `Number(delay) || 30` at server.js:4087 turns an explicit `delay: 0` (a caller asking for fast typing) into a 30ms jittered hold, and the sibling openclaw type path at server.js:6463 is untouched, so the two /type implementations now behave differently.
- ⚠️ `tests/unit/timeoutContinuation.test.js:27` - The new assertion `expect(clickRoute).toContain(&#39;withTimeout(_naturalMouseMove(&#39;)` (tests/unit/timeoutContinuation.test.js:27) only greps server.js source text for a token; it passes for commented-out or dead code and breaks on any behavior-preserving rename, proving nothing about timeout behavior. Replace with an executable check (e.g. drive _naturalMouseMove/_randomPointInBox against a fake page and assert the returned point stays inside the box and that a hung mouse.move rejects within the budget). Note there is currently no executable coverage at all for the new glide math.
- ℹ️ `server.js:3528` - _lastMousePos is module-level state shared by every session, tab and page (server.js:3528), but withTabLock only serialises per tab, so two users clicking concurrently each start their glide from the other&#39;s last coordinate on a different page. The click still lands exactly on target, so only path realism degrades; storing the cursor position on tabState would keep it per-page.
- ℹ️ `server.js:6398` - The parallel openclaw click path (server.js:6388-6407) still uses a plain locator.click() with no glide, so clicks routed through that endpoint keep the teleport signature. Out of the stated intent&#39;s scope (/click only); noted so the gap is a conscious one.

🔧 Fix: restore click error propagation, glide before actionable click
4 issues (2 warnings, 2 infos) still open:

- ⚠️ `server.js:3591` - The pre-click glide uses locator.boundingBox() coordinates without scrolling the element into view first, so for any target below the fold the cursor is eased toward a viewport coordinate that is off-screen. Concrete sequence: page scrolled to top, target button at document y=1800 in a 900px viewport -&gt; boundingBox() returns y~1800 -&gt; _naturalMouseMove walks the cursor to (x,1800), outside the viewport, dispatching mouse moves over nothing -&gt; locator.click() then does its own scrollIntoViewIfNeeded and clicks the element at its post-scroll position, which the cursor never traversed. The click still lands (correctness is fine), but the intent&#39;s stated goal (&#34;the cursor visibly eases/jitters through a multi-segment path to the target&#34;) is not achieved in exactly the common case, and a move-to-nowhere followed by a click elsewhere is itself an odd trace. Smallest remedy: await locator.scrollIntoViewIfNeeded({timeout: bboxTimeout}) inside moveToElement before boundingBox() (click() already scrolls, so net page behavior is unchanged).
- ⚠️ `lib/humanize.js:12` - randomPointInBox&#39;s padFrac parameter and the Math.min(box.width * padFrac, box.width * 0.4) clamp are components the intent does not require: the single caller (server.js:3550) never passes padFrac, and with the default 0.2 the clamp is unreachable because 0.2*w is always less than 0.4*w, so the min() is dead in every execution. The steps = 8 parameter of naturalMouseMove is likewise never overridden by any caller. Recommend removing the padFrac and steps parameters and inlining the 0.2 pad and 8 steps as constants, rather than keeping configurable knobs no requirement asks for.
- ℹ️ `server.js:3571` - On any fallback path, moveToElement runs twice for a single /click: once via glideToElement at server.js:3591 and again inside dispatchMouseSequence at server.js:3571, each doing a fresh boundingBox() plus a ~200-400ms jittered walk to a different random point. That costs an extra bounding-box resolution (up to 3s when the element detached) and an extra glide out of the same handler budget. Since the cursor is already parked inside the element by then, dispatchMouseSequence could reuse the point glideToElement landed on (tabState.mousePos) rather than re-resolving and re-walking. Harmless for correctness; noted as latency on the already-degraded path.
- ℹ️ `server.js:3552` - withTimeout(..., &#39;native mouse move&#39;) produces a message containing &#39;timed out after&#39;, which isTimeoutError() matches, and &#39;click&#39; is in NAVIGATION_TIMEOUT_ACTIONS (server.js:1499), so such an error destroys the entire user session for a fresh proxy. The label is pre-existing, but the wrapped operation grew from a single instantaneous mouse.move to a ~200-400ms multi-step walk, so exceeding a nearly-exhausted remainingBudget() is now materially more likely. In practice the outer HANDLER_TIMEOUT_MS tab destruction fires ~2s later anyway, so the blast radius is unchanged; noting the tradeoff rather than asking for a change.
</details>

<details>
<summary>⚠️ **Test** - 1 info</summary>

- 🚨 `lib/humanize.js:24` - With the product-default HANDLER_TIMEOUT_MS=30000 (lib/config.js:111), a single POST /tabs/:id/click on a trivial static local page now fails: the endpoint returns 410 {&#34;error&#34;:&#34;action timed out after 30000ms&#34;,&#34;code&#34;:&#34;tab_timeout&#34;,&#34;recovery&#34;:&#34;create_new_tab&#34;} and the tab is destroyed (a second run failed as &#39;native mouse move timed out after 1ms&#39;, i.e. budget already exhausted). Baseline e5a36f5 returns 200 in 8-10s on the same page/script. Cause: camoufox is launched with humanize:true (server.js:1125), so each page.mouse.move is itself expanded into a slow humanized micro-path; the new 8-step glide issues 9 such moves per click, taking the click from ~6s to ~25-27s (measured with the timeout raised to 60s: 25282/27076/27124ms, 611-645 mousemove events vs 111-113 at baseline). Reducing the step count, skipping the glide when camoufox humanize is already on, or budgeting the glide against remainingBudget()/step are all viable, so the call is the author&#39;s.
- ℹ️ `tests/unit/timeoutContinuation.test.js:39` - tests/unit/timeoutContinuation.test.js &gt; &#39;destroys a tab before releasing queued evaluate work after an evaluate timeout&#39; fails on this macOS host on BOTH the base commit and the target commit: the suite starts the server with HANDLER_TIMEOUT_MS=5000 and Camoufox cold start plus tab creation exceeds that, so createTab times out before the assertions run. Pre-existing environment sensitivity, not caused by this change.
- <code>`npx jest tests/unit/humanize.test.js` (10 tests, all pass - glide math, landing point, jitter bounds, error propagation)</code>
- <code>`npx jest tests/unit/clickBoundingBoxTimeout.test.js` (pass)</code>
- <code>`npx jest tests/unit/timeoutContinuation.test.js` (1 pre-existing env failure, reproduced identically on base e5a36f5)</code>
- <code>Manual end-to-end: `node glide-evidence.mjs` - starts the real camofox server + Camoufox browser, serves an instrumented page that records every mousemove and paints the cursor trail on a canvas, POSTs /tabs/:id/click {selector:&#39;#b&#39;}, reads back the trail and click count, captures a screenshot</code>
- <code>Same script run against base commit via `git checkout e5a36f5 -- server.js` for a before/after comparison (restored afterwards; worktree clean)</code>
- <code>Same script run at the product default `HANDLER_TIMEOUT_MS=30000` on both commits (2 runs each) to check the click still completes</code>

🔧 Fix: fix: single humanized glide move, drop redundant JS stepping
1 info still open:

- ℹ️ `tests/unit/timeoutContinuation.test.js:39` - tests/unit/timeoutContinuation.test.js &gt; &#39;destroys a tab before releasing queued evaluate work after an evaluate timeout&#39; still fails on this macOS host: the suite starts the server with HANDLER_TIMEOUT_MS=5000 and Camoufox cold start plus tab creation exceeds that, so createTab times out (422/410) before the assertions run. Confirmed in round 1 to fail identically on the base commit e5a36f5, so it is host-speed sensitivity, not a defect in this change. The two source-contract tests in the same file (including the one this change updated) pass.
- <code>`NODE_OPTIONS=&#39;--experimental-vm-modules&#39; npx jest --runInBand --forceExit tests/unit/humanize.test.js` (new executable coverage for randomPointInBox/rand)</code>
- `NODE_OPTIONS='--experimental-vm-modules' npx jest --runInBand --forceExit tests/unit/clickBoundingBoxTimeout.test.js tests/unit/timeoutContinuation.test.js`
- <code>Manual end-to-end: `node glide-evidence.mjs` (x2) - real camofox server + Camoufox tab, local page records every mousemove, POST /tabs/:id/click on a button, then the recorded path is drawn back onto the page and screenshotted</code>
- <code>Manual end-to-end: `node missing-target-evidence.mjs` - /click on a non-existent selector and on a display:none element, asserting the request still fails (422) rather than reporting ok:true</code>
- <code>Manual end-to-end: `node blocked-click-evidence.mjs` - /click on a button covered by a full-screen overlay, plus the same click after removing the overlay</code>
- `Timing comparison against the round-1 baseline measurements at HANDLER_TIMEOUT_MS=30000 (base e5a36f5 vs previous head e803440 vs current 668c2f3)`
</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.
</details>


Wilsonchoo2017:
Closing: this work is landing in the Wilsonchoo2017/camofox-browser fork instead of upstream.

